PR #10, opened 2015-08-14
Hello

schmaltzlicker:
Raw memory iterator proposal
destruct algorithm
unstable remove algorithms and tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,21 @@ $RECYCLE.BIN/
 Network Trash Folder
 Temporary Items
 .apdisk
+*.tlog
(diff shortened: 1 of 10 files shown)


johnmcfarlane:
Mind if I rearrange the files a bit? I'd like to move source files, unstable_remove_test.cpp and main.cpp, from /VS2015/SG14/SG14_test/ to /SG14/SG14_test where the cmake project can more easily see them. Then I'll remove the duplicate main.cpp in /cmake/. 

This seems like an obvious clean-up but I don't want a vexing merge to jump out at you if you're actively working in these files!

schmaltzlicker:
OK w/ me

PR #18, opened 2015-08-23
Plf cmake fixes

johnmcfarlane:
I hit a few hitches - as expected - when compiling new additions to SG14_test under the cmake system. As I had the relevant build environment to hand and the changes looked straightforward, I figured it easiest to just go ahead and submit a branch with proposed fixes. This PR should be reviewed by the original author of the code, @mattreecebentley. Hope it's OK! (The tests seem to pass BTW.)


mattreecebentley:
All good - I never knew const was ignored on return type - never got the warning from GCC! Good to know, will update in master copy.

Thanks :)

johnmcfarlane:
Glad to help.

As I understand it, because the function is returning by value, a copy of
the object effectively gets made in the calling function. And copying an
object doesn't copy its `const`ness. It's equivalent to:

    const int returned_value = 1;
    int caller_copy = returned_value;  // for this line to compile, it
doesn't matter whether returned_value is const or not

Now why the compiler should care is another matter!

On Sat, Aug 22, 2015 at 11:58 PM, mattreecebentley <notifications@github.com

mattreecebentley:
Sure - that makes sense.
Only place it should have an effect I guess is reference returns.

mattreecebentley:
Sure - that makes sense.
Only place it should have an effect I guess is on reference returns.

WG21-SG14:
Hello chaps.  I'm on holiday at the moment and only have occasional internet access.  I'll be back on September 2nd, so excuse any slow service.  Thanks for staying on top of this hitch.

const return value signatures are particularly useful for overloading square operator where you might want a construct like this:

const my_container<foo>& c = get_container();
c[6].func();

If func is a non-const member function, this should fail.

Cheers,
Guy

WG21-SG14:
Mmmm, formatting not your friend here.  Assume my_container is an alias of a container over some type foo.

mattreecebentley:
Minor mistake, so to save me doing a full fork/pull/etc request series can someone please change the following line in plf::colony from:

class iterator : public std::iterator<std::bidirectional_iterator_tag, element_type, std::ptrdiff_t, element_type *, element_type &>

to

class iterator : public std::iterator<std::random_access_iterator_tag, element_type, std::ptrdiff_t, element_type *, element_type &>

mattreecebentley:
(missing text due to comment box interpreting code as HTML fixed, see comment on website if you received comment on email)

PR #47, opened 2015-09-27
Swardle branch

swardle:
Add cpp_entity_example


swardle:
Creating some examples that show the performance of comparing things to vtables.

PR #54, opened 2015-10-24
Sg14 from swardle master compare vtable function

swardle:
Update ComparingVirtualMethods to include problem with comdat folding


swardle:
Update ComparingVirtualMethods to include problem with comdat folding

PR #61, opened 2016-04-28
colony v3.05 - Removal of compiler-specific iteration code, resulting…

mattreecebentley:
… in 8% speedup for larger-than-scalar type iteration, 6% slowdown for scalar-type iteration, under GCC x64. MSVC unaffected by change. Correction to plf_nanotimer under linux/bsd.

Benchmarks completely rewritten, graphs of all results against std:: library containers and container modifications in docs. Benchmarks separated into benchmark suite.
Corrections to reinitialize asserts. Added bound-checking asserts to advance/next/prev implementations.


mattreecebentley:
Various updates. Full documentation now includes detailed graphs against other std:: containers, for various numbers of elements (10-100000) and storage types (char, int, double, small struct, large struct).

PR #70, opened 2016-07-20
2016-07-19: v3.5

mattreecebentley:
2016-07-19: v3.5 - Colony now accepts a template parameter for the skipfield type. This enables the use of, for example, unsigned char as the skipfield type. The bitdepth of the unsigned integer also defines the default maximum size of colony groups. The benchmark suite has been updated with memory measurements for the general_use tests. General_use tests changed to percentage-based tests. Internal erased location stack now stores only element pointers, not full iterators. Resultant code is faster and consumes less memory. Slight improvement for fill-insert. Sanity-checks for range-insert and initializer-list-insert implemented. Minor fix to C++03 support in benchmark suite.
2016-07-4: v3.35 - Another compliance fix for benchmark suite in clang, courtesy of Arthur O'Dwyer.
2016-07-3: v3.34 - Minor warning removal for test suite. Performance improvement for colony edge-case scenario where a single element is inserted and then erased repeatedly, and where the insertion point happens to be at the end of a memory block.
2016-07-1: v3.33 - Empty destructors/constructors removed in iterators. Small compliancy fixes in both colony and benchmark suite. Thanks to Arthur O'Dwyer for this update.
2016-06-25: v3.31 - Colony iterators no longer inherit from std::iterator and implement the required typedefs themselves. Code tidyup and minor changes. Default of 1 implemented for prev/next.
2016-06-17: v3.30 - Correction to colony range constructor. Corrections to advance() asserts.
2016-06-15: v3.29 - Support for stateful allocators added, to both colony and stack. More generic solution applied for preventing calls for fill-insert to be matched by range insert template.


carlcook:
Massive change!!

carlcook:
Hey Matt, would you like to be a committer to SG14, i.e. part of the group? This way you can merge pull requests yourself.

mattreecebentley:
Would love to - would certainly make things easier for all involved.

carlcook:
Hi Matt, great, I've added you as a member of the SG14 team, with write permissions to the SG14 repo.

Guy (et. al.), I've also just locked the default permissions a bit, so that now SG14 team members get write permission to the SG14 repo, instead of admin permission to the SG14 team (i.e. only the WG21 admins can add/remove new repos, team members, etc). I think this makes good sense!

mattreecebentley:
Cheers guys!

PR #71, opened 2016-10-03
Fix 2 moving functions to not be forward/universal references.

nidoizo:
Might want to submit fix before my talk is on Youtube @WG21-SG14/sg14.


carlcook:
Looks good to me

Issue #19, opened 2015-08-23
Invalid platform toolset?


briterator:
My copy of Visual Studio 2015  does not accept the toolchain specified in the vcxproj files. It states that the given toolset is not installed. If I change the toolset back to just v140 as it used to be instead of Visual Studio 2015 (v140) then it works properly.

johnmcfarlane:
I'm not sure how, but I think this was probably me. I noticed this earlier
but have a bad install and am half way through reinstalling Windows now. If
it's not fixed by the time I'm back up, I'll take care of it. 😖

WG21-SG14:
I'm away from my PC for another week.  Is one of you able to close this issue now?

Cheers,
Guy

johnmcfarlane:
I'm not seeing this error in master any more. Still need to fix it in my
current branch though for some reason. BTW, any chance of being added to
the project?
Cheers
John

PR #82, opened 2016-11-07
Removing VS2015 files. Generate them using CMake...

davidziman:
Removing VS2015 files. Generate them using CMake. Stale ring_test.cpp file has been removed to avoid confusion.


davidziman:
This should be sufficient to close issue #19.

carlcook:
Thanks for this, David. If the ring tests ever need to be resurrected, they will be in the commit history.

PR #83, opened 2017-02-01
Added Ring Proposal R4

hatcat:


mattreecebentley:
A Few minor comments: what's the rationale for why iterators aren't
appropriate for non-owned memory? Also for dunderheads like myself, what
owns the memory in this case?
In the last section, a minor mistake: 'the a fixed size...'.
lastly, there should be no connection between 'non-contiguous' containers
and triggering an allocation in Motivation section, as even contiguous
containers like vector may trigger alloc/dealloc on insert.
Otherwise reads well.

hatcat:
Thanks Matt. To be honest, I'm not entirely sure of the no-iterator rationale, it was passed to me by Michael Wong at Meeting C++ during a noisy post-session social so my recall is imperfect. I would welcome a clear explanation; possibly it's to do with invalidation by the owner.

mattreecebentley:
Gotcha - I guess my counter to that would be the same counter I got from 
a committee member (I forget who) for my replacement stack proposal for 
STL2- even if you're not likely to use it, it's useful for debugging 
purposes.
Probably something that needs a discussion on forums or something.

PR #86, opened 2017-04-13
Fix various ring_iterator snafus and a bug in push_back.

Quuxplusone:
Fix several bugs reported by @paulbendixen on the mailing list:

- Runtime misbehavior in `operator==` and `operator<` (e.g. `for` each element of a full container was unintentionally a no-op)
- Lack of `operator!=`
- Lack of `operator>` `operator>=` `operator<=`
- Lack of `operator+` `operator-`
- Compile error in `cbegin()`
- Runtime misbehavior in `increase_size()` (e.g. as called from `push_back()` on a full container)

(I'm happy to have these four commits cherry-picked in, or merged, or squash-merged; doesn't much matter to me.)

### DIFF
--- a/SG14/ring.h
+++ b/SG14/ring.h
@@ -132,7 +132,6 @@ namespace sg14
 		friend Ring;
 		using size_type = typename Ring::size_type;
 		ring_iterator(size_type idx, std::conditional_t<is_const, const Ring, Ring>* rv) noexcept;
-		size_type modulo_capacity(size_type idx) const noexcept;
 		size_type m_idx;
 		std::conditional_t<is_const, const Ring, Ring>* m_rv;
 	};
@@ -344,7 +343,7 @@ template <typename Ring, bool is_const>
 template<bool C>
 bool sg14::ring_iterator<Ring, is_const>::operator==(const sg14::ring_iterator<Ring, C>& rhs) const noexcept
 {
-	return (modulo_capacity(m_idx) == rhs.modulo_capacity(m_idx)) && (m_rv == rhs.m_rv);
+	return (m_idx == rhs.m_idx) && (m_rv == rhs.m_rv);
 }
 
 template <typename Ring, bool is_const>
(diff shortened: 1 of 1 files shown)


mattreecebentley:
Does < or > even make sense in a ring?

SG14/ring.h
```diff
@@ -108,7 +108,15 @@ namespace sg14
 		template <bool C>
 		bool operator==(const ring_iterator<Ring, C>& rhs) const noexcept;
 		template <bool C>
+		bool operator!=(const ring_iterator<Ring, C>& rhs) const noexcept;
+		template <bool C>
 		bool operator<(const ring_iterator<Ring, C>& rhs) const noexcept;
```
Quuxplusone: @mattreecebentley commented:
> Does < or > even make sense in a ring?

I believe it does. `x < y` should be true iff `x - y < 0` is true, and I suppose you'd agree that `x - y` ought to be well-defined in a ring, right?

An alternative approach would be to give up on RandomAccessIterator and just have ring_iterator be a BidirectionalIterator, in which case we could omit the < <= > >= comparison operators; but I don't see any reason we *should* do that. RandomAccessIterator seems like the natural category to me.

mattreecebentley: I don't agree that that is the case, no. x is always going to be both > and < of y in a ring iterator.

Quuxplusone: Hm. So in a ringbuffer, you'd say it was *not* necessarily true that `rv.end() - rv.begin() == rv.size()`?

mattreecebentley: I guess it's all semantics. In my view comparison operations make no sense because that doesn't match a ring. In practical terms usage may be different. I'll let the actual authors, who're more experienced with their use, puzzle it out with you.

hatcat:
Thanks so much for your input to the ring, but unfortunately I've just noticed that this part of the repo is out of sync with the paper. In R4 of the paper we removed iterators entirely after feedback from Issaquah; you can see this https://github.com/WG21-SG14/SG14/blob/master/Docs/Proposals/RingProposal.pdf which is up to date. Apologies again, I do abhor wasting people's time. I will update the ring as soon as possible.

Quuxplusone:
@hatcat: FWIW, I recommend merging this PR anyway, and then going in *afterward* to rip out the iterators if you really still want to. (My impression was that feedback from the most recent (Kona?) meeting had included "hey where did the iterators go?". Besides, you already know that I consider ringbuffer 99% useless without iteration. By the way, this is Arthur. ;))

The bugfix to `.increase_size()` is necessary either way, but could certainly be cherry-picked on its own.

hatcat:
Hi Arthur.
That IS perplexing. The feedback I had from Issaquah was that there should be no iterators. I'm hoping to attend Toronto and just have it out with everyone. I can't help feeling we're being dicked around by whoever happens to be there on the day: the feedback does appear to have been contradictory from session to session. I'm going to check on the feedback posted to the reflector and add it to the repo. Then I'll merge &c.

Perhaps the winning strategy is to present something finished, wording and all, and simply bludgeon them into acceptance! The unfinished state gives scope for tweaking.

paulbendixen:
Great, I've updated my [test program](https://github.com/paulbendixen/testring) to show a reason why iterators would make very much sense in the way I see it.

Quuxplusone:
By the way @paulbendixen, I noticed after making this patch that there's a test file in *this* repo under `SG14_test/ring_test.cpp`. I suggest that you fork this repo, add your test program as a new test case in that file, and submit a pull request. (Honestly I *should* have added tests when making *this* PR; but I didn't see the tests at the time due to laziness, and now that you're basically maintaining a test suite off to the side, I'd think it would make sense for you to merge it in.)

(Whereas, if I went down that rabbit hole myself, I'd probably spend even more time trying to make `CXXFLAGS='-std=c++14 -I../SG14' make ring_test` Do The Right Thing out of the box than I just spent trying to figure out the real build structure.)

paulbendixen:
@Quuxplusone Good idea, I'll look into it, how ever I seem to have just about just the same luck compiling the entire SG14 project.
I don't feel good making a PR without being able to compile it.
Maybe it's time to look into conan.io

PR #87, opened 2017-04-15
`#pragma once` all the things!

### LINKED ISSUE #75: add include guards (or pragma once) to ring and inplace_function headers

Quuxplusone:
After this change,

    find . -name '*.h' | xargs head -1

shows that every .h file in the repo correctly begins with `#pragma once`.
A few of them have include guards *also*, so the style isn't completely
uniform; but at least none of them are completely unprotected anymore.

Fixes #75.

### DIFF
--- a/Proposals/ComparingVirtualMethods/cpp_entity_example/stdafx.h
+++ b/Proposals/ComparingVirtualMethods/cpp_entity_example/stdafx.h
@@ -1,9 +1,9 @@
+#pragma once
+
 // stdafx.h : include file for standard system include files,
 // or project specific include files that are used frequently, but
 // are changed infrequently
 //
-
-#pragma once
 
 #ifdef _WIN32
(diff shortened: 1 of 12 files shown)


mattreecebentley:
Given that pragma once is (a) non-standard and unsupported by some lesser compilers, for example some embedded compilers (b) irrelevant in gcc for a decade and (c) yields no known compile time benefits on any current version of the major four, I'm closing this.

paulbendixen:
So, just to clarify: another PR that used ifdef include guards would be allowed?

mattreecebentley:
I think so - but some of that stuff doesn't actually need include guards - like the plf_bench, which is only meant to be included once. Individual authors can make up their own minds as to where they stand on the pragma once vs include guard thing - but I'm just closing it as much as it pertains to mine.

PR #88, opened 2017-04-15
Add a signal processing relevant test case

paulbendixen:
After suggestion from @Quuxplusone, added the tests I've shown in my own repo.

### DIFF
--- a/SG14_test/ring_test.cpp
+++ b/SG14_test/ring_test.cpp
@@ -6,6 +6,7 @@
 #include <mutex>
 #include <future>
 #include <iostream>
+#include <numeric>
 
 void sg14_test::ring_test()
 {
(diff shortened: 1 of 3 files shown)


SG14_test/ring_test.cpp
```diff
@@ -88,4 +89,26 @@ void sg14_test::thread_communication_test()
...
+	buffer.push_back( 5.0 );
+
+	assert( buffer.front() == 2.0 );
+
+	constexpr std::array< double, 3 > filter_coefficients = { 0.25, 0.5, 0.25 };
+
+	// In an update loop, interrupt routine or the like
+	buffer.push_back( 7.0 );
+
+	assert( std::inner_product( buffer.begin(), buffer.end(), filter_coefficients.begin(), 0.0 ) / buffer.size() == 20/3 );
```
Quuxplusone: Should this be simply `... == 6.0`? I think all the math is being done as floating-point, except for the `20/3` on the right.

paulbendixen: See my updated PR, I adopted the int version, but forgot to remove the division, You wouldn't want to do that in a filter.
Also the code was never run, because I forgot to add it to the testcase (Let he who is without sin...)

PR #97, opened 2017-05-06
Moved P0130 Comparing Virtual Functions

TBBle:
This paper has now been moved to its own repository in the WG21-SG14
group at https://github.com/WG21-SG14/SG14-comparing-virtual-functions


TBBle:
Since I've already had consensus from anyone I could think to ask, I'll accept this PR myself next week, or next time I think about it, whichever is the longer.

PR #95, opened 2017-05-06
SG14_test build fixes

TBBle:
A quick set of build-fixes for the SG14_test code and CMake build system.

Tested with:
* MSVC2015
* MSVC2017
* gcc 6.3.0 (mingw64 from MSYS2)
* clang 3.9.1 (mingw64 from MSYS2)

This doesn't fix the builds completely, but other fixes need to be done in the SG14 library itself, and I want to keep those separated, and ideally unit-tested.

PR #106, opened 2017-05-18
Suppress 'unused parameter' warnings for plf test suite

TBBle:
By some magic, this also suppresses a couple of other warnings in MSVC, triggered by the plf colony test suite.
* C4456: declaration of 'the_iterator' hides previous local declaration
* C4267: 'initializing': conversion from 'size_t' to 'unsigned short', possible loss of data

I _suspect_ these other warnings are normally suppressed unless another warning in the same file is triggered.

Broken out of #95 due to allow further discussion.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -32,6 +32,17 @@ set(SOURCE_FILES
     ${SG14_TEST_SOURCE_DIRECTORY}/transcode_test.cpp
     ${SG14_TEST_SOURCE_DIRECTORY}/uninitialized.cpp)
 
+if(MSVC)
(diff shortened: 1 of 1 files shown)


TBBle:
This pull request is the alternative to https://github.com/TBBle/SG14/tree/SG14_test-build-fixes_plf

TBBle:
Turns out the other two warnings do need to be suppressed, but it was masked by a different failure in an SG14 library. I've suppressed the other test-case warning (C4456) in PR #109. C4267 is occurring in the actual plf::colony header, and it appears to be in implicit code, so I'm still investigating.

mattreecebentley:
This should be sorted now. MSVC can get a little pedantic, that's all.

PR #108, opened 2017-05-19
Fix behavior of ring_span's copy_popper<T>, and general popper cleanup.

### LINKED ISSUE #91: Fix return of ring_span's copy_popper<T>::operator()
in:
```
template <typename T>
T sg14::copy_popper<T>::operator()(T& t)
{
	T old = t;
	t = copy;
	return t;
}
```
shouldn't `return t` be **`return old`**?

### LINKED ISSUE #102: Add missing copy_popper(T const&) ctor for ring_span
Missing from both proposal and reference code.

Quuxplusone:
Fixes #91. Apparently the typo `t` for `old` is also in the current
paper http://wg21.link/p0059r3 — we should fix that too!

Fixes #102.

### DIFF
--- a/SG14/ring.h
+++ b/SG14/ring.h
@@ -10,21 +10,22 @@ namespace sg14
 	template <typename T>
 	struct null_popper
 	{
-		void operator()(T&);
+		void operator()(T&) const noexcept;
 	};
 
 	template <typename T>
 	struct default_popper
 	{
-		T operator()(T& t);
+		T operator()(T& t) const;
 	};
 
 	template <typename T>
 	struct copy_popper
 	{
-		copy_popper(T&& t);
-		T operator()(T& t);
-		T copy;
+		explicit copy_popper(T t);
+		T operator()(T& t) const;
+	private:
+		T m_copy;
 	};
 
 	template <typename, bool>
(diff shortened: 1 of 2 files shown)


SG14/ring.h
```diff
@@ -157,26 +158,27 @@ namespace sg14
...
-T sg14::copy_popper<T>::operator()(T& t)
+T sg14::copy_popper<T>::operator()(T& t) const
 {
-	T old = t;
-	t = copy;
-	return t;
+	T old = m_copy;
+	using std::swap;
+	swap(old, t);
+	return old;
```
Quuxplusone: This dance makes the operation exception-safe, in case that matters. If copying `m_copy` into `old` throws an exception, then the swap doesn't happen and the value of `t` is preserved.

SG14/ring.h
```diff
@@ -10,21 +10,22 @@ namespace sg14
...
+		T operator()(T& t) const;
 	};
 
 	template <typename T>
 	struct copy_popper
 	{
-		copy_popper(T&& t);
-		T operator()(T& t);
-		T copy;
+		explicit copy_popper(T t);
```
Quuxplusone: Notice that I'm changing this from `T&&` to just `T`, which fixes #102 in an elegant way. I don't think there's any downside to this change.

TBBle: The `explicit` broke one of the tests included here: https://github.com/WG21-SG14/SG14/commit/aba97661a03d99712d87bba61c1d4fd03569eb05#commitcomment-22430301

PR #109, opened 2017-05-19
Suppress another warning in plf_colony test suite

TBBle:
This suppresses:
* C4456: declaration of 'the_iterator' hides previous local declaration

mentioned in cf337d3a4361a519e831bb812915d78aa46c6959. It came back after fixing another error, looks like Visual Studio was failing-fast.


TBBle:
This PR is the alternative to 8fc79e9487d7400ae486e2a635de8ffa036bbbac. Personally, I prefer the other change over the warning suppression.

This one only affects MSVC (gcc doesn't complain even with `-Wshadow`) so I'm not in a huge hurry to resolve it.

mattreecebentley:
Incidentally, I think this was fixed with my last push. But as to whether or not it's still required to suppress the other warnings that MSVC brings up, like use of uninitialized_copy I don't know.

PR #110, opened 2017-07-31
Missing Alignment Parameter Fix

jumonatr:
- added required template parameter for alignment when constructing and assigning inplace_functions
- added a static_assert validating alignment for those two cases

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -101,8 +101,8 @@ public:
 	// Allows for copying from inplace_function object of the same type, but with a smaller buffer
 	// May throw any exception encountered by the constructor when copying the target object
 	// If OtherCapacity is greater than Capacity, a compile-time error is issued
-	template<size_t OtherCapacityT>
-	inplace_function(const inplace_function<RetT(ArgsT...), OtherCapacityT>& other)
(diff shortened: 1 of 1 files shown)


SG14/inplace_function.h
```diff
@@ -224,10 +224,11 @@ class alignas(AlignmentT) inplace_function<RetT(ArgsT...), CapacityT, AlignmentT
 		m_ManagerFctPtr = nullptr;
 	}
 
-	template<size_t OtherCapacityT>
-	void copy(const inplace_function<RetT(ArgsT...), OtherCapacityT>& other)
+	template<size_t OtherCapacityT, size_t OtherAlignment>
```
Quuxplusone: LGTM :shipit: — but here and throughout, you might want to rename `OtherCapacityT` to just `OtherCapacity`. Looks like the existing code is inconsistent in how it names this parameter... but it's totally not a `T`ype.

jumonatr: I agree, I just didn't want to batch that change with this one and wanted to keep it consistent. Maybe in a follow up PR.

carlcook:
Thanks guys!

Issue #111, opened 2017-09-14
inplace_function copy-assignment fails


Quuxplusone:
https://wandbox.org/permlink/nT7NOhnVm2JZopjh

    stdext::inplace_function<void(), 16> f;
    f = []() { puts("hello world"); };
    stdext::inplace_function<void(), 16> g;  // bonus: change 16 to 17 and it should still work
    g = f;  // ERROR!

Oddly, it *does* work for move-assignment!
I think the issue is some SFINAE on the wrong overload. Should be:

	template<typename Callable, class = typename std::enable_if<!std::is_lvalue_reference<Callable>::value>::type>
	inplace_function& operator=(Callable&& target)

That SFINAE looks sketchy to me anyway, but it seems to DTRT.

carlcook:
Thanks! I'll look at this soon

carlcook:
Yeah you are right. Here's the fix I think.

```
template<typename CallableT>
inplace_function& operator=(const CallableT& target) {

template<typename CallableT, class = typename std::enable_if<!std::is_lvalue_reference<CallableT>::value>::type>
inplace_function& operator=(CallableT&& target)
```

And indeed, without this change, copy assignment is failing, but move assignment still works.

carlcook:
I still haven't committed this fix, but will do in a few days (I'm currently on leave and don't have workstation access...)

carlcook:
Looks like Arthur made some changes, including this fix.

PR #112, opened 2017-10-03
Bugfixes and a better test harness for stdext::inplace_function

Quuxplusone:
This includes a mass renaming of template parameters so that all the type parameters are named "FooT" and all the non-type parameters are named "Foo" (no "T").

This also includes the fix for #111.

However, I've found another problem with the conversion from `inplace_function<Sig, Sz>` to `inplace_function<Sig, LargerSz>` — namely, if the source is the "empty" function object, then the destination receives a function object that it doesn't recognize as "empty".

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -42,21 +42,21 @@ enum class inplace_function_operation
 	Move
 };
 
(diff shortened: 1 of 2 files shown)


Quuxplusone:
...And now I've fixed the "empty function object conversion" bug, in a branch based on top of this one. See https://github.com/WG21-SG14/SG14/compare/master...Quuxplusone:inplace-function-2 and feel free to review/pull that one as well.

PR #116, opened 2017-11-13
Rework inplace_function implementation

Voultapher:
Safer and faster implementation
No function pointer reinterpret_cast
static_cast over reinterpret_cast to facilitate portability

Passes testsuit.
Haven't done extensive benchmarks, however [this](http://quick-bench.com/O0DT4eaK38jvVBlnHLg4jXB_LhA) seems promising.

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -79,10 +79,10 @@ template<typename R, typename... Args> struct vtable
 
   template<typename C> explicit constexpr vtable(wrapper<C>) noexcept :
     invoke_ptr{ [](storage_ptr_t storage_ptr, Args&&... args) -> R
-			{ return (*static_cast<C*>(storage_ptr))(
+      { return (*static_cast<C*>(storage_ptr))(
         std::forward<Args>(args)...
       ); }
-		},
+    },
 		copy_ptr{ [](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
 		  { new (dst_ptr) C{ (*static_cast<C*>(src_ptr)) }; }
     },
@@ -104,11 +104,7 @@ template<typename R, typename... Args> struct vtable
 };
 
 template<typename R, typename... Args>
-static const vtable<R, Args...> empty_vtable{};
-
-template<typename R, typename... Args>
-static const auto empty_vtable_ptr =
-std::addressof(empty_vtable<R, Args...>);
+inline constexpr vtable<R, Args...> empty_vtable{};
 
 template<size_t DstCap, size_t DstAlign, size_t SrcCap, size_t SrcAlign>
 struct is_valid_inplace_dst : std::true_type
@@ -150,7 +146,7 @@ public:
   template <typename, size_t, size_t>	friend class inplace_function;
 
   inplace_function() noexcept :
-    vtable_ptr_{detail::empty_vtable_ptr<R, Args...>}
+    vtable_ptr_{std::addressof(detail::empty_vtable<R, Args...>)}
   {}
 
   template<
@@ -186,7 +182,7 @@ public:
   }
 
   inplace_function(std::nullptr_t) noexcept :
-    vtable_ptr_{detail::empty_vtable_ptr<R, Args...>}
+    vtable_ptr_{std::addressof(detail::empty_vtable<R, Args...>)}
   {}
 
   inplace_function(const inplace_function& other) :
@@ -205,13 +201,13 @@ public:
       std::addressof(storage_),
       std::addressof(other.storage_)
     );
-    other.vtable_ptr_ = detail::empty_vtable_ptr<R, Args...>;
+    other.vtable_ptr_ = std::addressof(detail::empty_vtable<R, Args...>);
   }
 
   inplace_function& operator= (std::nullptr_t) noexcept
   {
     vtable_ptr_->destructor_ptr(std::addressof(storage_));
-    vtable_ptr_ = detail::empty_vtable_ptr<R, Args...>;
+    vtable_ptr_ = std::addressof(detail::empty_vtable<R, Args...>);
     return *this;
   }
 
@@ -242,7 +238,7 @@ public:
         std::addressof(other.storage_)
       );
 
-      other.vtable_ptr_ = detail::empty_vtable_ptr<R, Args...>;
+      other.vtable_ptr_ = std::addressof(detail::empty_vtable<R, Args...>);
     }
     return *this;
   }
@@ -272,7 +268,7 @@ public:
 
   explicit constexpr operator bool() const noexcept
 	{
-	  return vtable_ptr_ != detail::empty_vtable_ptr<R, Args...>;
+	  return vtable_ptr_ != std::addressof(detail::empty_vtable<R, Args...>);
 	}
 
   template<size_t Cap, size_t Align>
@@ -340,7 +336,7 @@ private:
       storage_ptr
     );
 
-    *vtable_ptr = detail::empty_vtable_ptr<R, Args...>;
+    *vtable_ptr = std::addressof(detail::empty_vtable<R, Args...>);
   }
 };
 
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -69,40 +69,29 @@ template<typename R, typename... Args> struct vtable
   const destructor_ptr_t destructor_ptr;
 
   explicit constexpr vtable() noexcept :
-    invoke_ptr{ static_cast<invoke_ptr_t>(
-      [](storage_ptr_t, Args&&...) -> R
+    invoke_ptr{ [](storage_ptr_t, Args&&...) -> R
       { throw std::bad_function_call(); }
-    )},
-    copy_ptr{ static_cast<copy_ptr_t>(
-      [](storage_ptr_t, storage_ptr_t) noexcept -> void {}
-    )},
-    move_ptr{ static_cast<copy_ptr_t>(
-      [](storage_ptr_t, storage_ptr_t) noexcept -> void {}
-    )},
-    destructor_ptr{static_cast<destructor_ptr_t>(
-      [](storage_ptr_t) noexcept -> void {}
-    )}
+    },
+    copy_ptr{ [](storage_ptr_t, storage_ptr_t) noexcept -> void {} },
+    move_ptr{ [](storage_ptr_t, storage_ptr_t) noexcept -> void {} },
+    destructor_ptr{ [](storage_ptr_t) noexcept -> void {} }
   {}
 
   template<typename C> explicit constexpr vtable(wrapper<C>) noexcept :
-    invoke_ptr{ static_cast<invoke_ptr_t>(
-      [](storage_ptr_t storage_ptr, Args&&... args) -> R
+    invoke_ptr{ [](storage_ptr_t storage_ptr, Args&&... args) -> R
 			{ return (*static_cast<C*>(storage_ptr))(
         std::forward<Args>(args)...
       ); }
-		)},
-		copy_ptr{ static_cast<copy_ptr_t>(
-      [](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
+		},
+		copy_ptr{ [](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
 		  { new (dst_ptr) C{ (*static_cast<C*>(src_ptr)) }; }
-    )},
-    move_ptr{ static_cast<copy_ptr_t>(
-      [](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
+    },
+    move_ptr{ [](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
 		  { new (dst_ptr) C{ std::move(*static_cast<C*>(src_ptr)) }; }
-    )},
-		destructor_ptr{ static_cast<destructor_ptr_t>(
-			[](storage_ptr_t storage_ptr) noexcept -> void
+    },
+		destructor_ptr{ [](storage_ptr_t storage_ptr) noexcept -> void
 			{ static_cast<C*>(storage_ptr)->~C(); }
-		)}
+		}
   {}
 
   vtable(const vtable&) = delete;
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -287,21 +287,24 @@ public:
 	}
 
   template<size_t Cap, size_t Align>
-  operator inplace_function<R(Args...), Cap, Align>() const
+  operator inplace_function<R(Args...), Cap, Align>() const&
   {
     static_assert(detail::is_valid_inplace_dst<
       Cap, Align, Capacity, Alignment
     >::value);
 
-    std::aligned_storage_t<Cap, Align> conv_storage;
-    vtable_ptr_->copy_ptr(
-      std::addressof(conv_storage),
-      std::addressof(storage_)
-    );
-    return {vtable_ptr_, std::move(conv_storage)};
+    return {vtable_ptr_, std::addressof(storage_)};
   }
 
-  // not sure a portable memcopy && conversion solution exists
+  template<size_t Cap, size_t Align>
+  operator inplace_function<R(Args...), Cap, Align>() &&
+  {
+    static_assert(detail::is_valid_inplace_dst<
+      Cap, Align, Capacity, Alignment
+    >::value);
+
+    return {std::addressof(vtable_ptr_), std::addressof(storage_)};
+  }
 
   void swap(inplace_function& other)
   {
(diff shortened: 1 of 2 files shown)


SG14/inplace_function.h
```diff
@@ -29,340 +29,308 @@
...
+  }
+
+  inplace_function(inplace_function&& other) :
+    vtable_ptr_{other.vtable_ptr_}
+  {
+    vtable_ptr_->move_ptr(
+      std::addressof(storage_),
+      std::addressof(other.storage_)
+    );
+    other.vtable_ptr_ = detail::empty_vtable_ptr<R, Args...>;
```
Quuxplusone: And also `vtable_ptr_->destructor_ptr(std::addressof(other.storage_))`, yeah? Otherwise you have an undestroyed contained-object in `other` but you've lost the vtable_ptr that knows how to destroy it.
I'd like to add a unit test that checks this case; I should get to it later today if I don't hear anyone objecting to the idea.

SG14/inplace_function.h
```diff
@@ -29,340 +29,308 @@
...
-		case Operation::Move:
-			{
-				FunctorT* source = (FunctorT*)fromPtr;
-				new (thisFunctor) FunctorT(std::move(*source));
-				break;
-			}
-		}
-	}
+  inplace_function(vtable_ptr_t vtable_ptr, storage_t&& storage) noexcept :
+    vtable_ptr_{vtable_ptr}, storage_{storage}
```
Quuxplusone: I don't understand why `storage` is an rvalue reference here.
I am unsure of whether `storage_{storage}` is guaranteed to do something-akin-to-memcpy or whether it's allowed to be a compiler error. Brace-initializing `aligned_storage_t` is an idiom I haven't seen before.

SG14/inplace_function.h
```diff
@@ -29,340 +29,308 @@
...
+  void swap(inplace_function& other)
+  {
+    if (this == std::addressof(other)) return;
+
+    storage_t tmp;
+    vtable_ptr_->move_ptr(
+      std::addressof(tmp),
+      std::addressof(storage_)
+    );
+    vtable_ptr_->move_ptr(
```
Quuxplusone: I am fairly confident that at least one of these `vtable_ptr_`s should be an `other.vtable_ptr_`.

SG14/inplace_function.h
Quuxplusone: Pushed the new test in 6e4044fbf2812b09cc2f75e9d0b19f8b40be889f. As expected, the test is green on master and red on Voultapher's branch. (But the fix is trivial.)

SG14/inplace_function.h
```diff
@@ -29,340 +29,308 @@
...
+
+  explicit constexpr vtable() noexcept :
+    invoke_ptr{ static_cast<invoke_ptr_t>(
+      [](storage_ptr_t, Args&&...) -> R
+      { throw std::bad_function_call(); }
+    )},
+    copy_ptr{ static_cast<copy_ptr_t>(
+      [](storage_ptr_t, storage_ptr_t) noexcept -> void {}
+    )},
+    move_ptr{ static_cast<copy_ptr_t>(
```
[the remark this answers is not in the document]
Voultapher: Nothing it is

SG14/inplace_function.h
Voultapher: Shouldn't the moved from object in this case `other`, be in a state resembling destruction?

SG14/inplace_function.h
Voultapher: Fixed with direct inplace construction and copy, move awareness.

SG14/inplace_function.h
Quuxplusone: Test case filed as #117. It turns out that master is also broken here, but in a different (and possibly worse) way; #117 contains a fix for master as well as the new test case.

SG14/inplace_function.h
Quuxplusone: "Resembling" maybe, but definitely not "equal to", unless the destructor is trivial. I still don't understand the rules around when it's okay to skip a trivial constructor or destructor, but I *know* it's not allowed to skip a *non-*trivial constructor or destructor!

Voultapher:
[Updated benchmark](http://quick-bench.com/evEuxq8-kTTwY1Vx-GPIezXUM0w)

Some of the used features now mandate proper C++17 language and library support.

Seeing that this is a proposal for an upcoming standard, this shouldn't be too bad.
Maybe we should supply a version that supports older compilers, imo this shouldn't change the code to drastically. Still I'd like the current version to keep its advantages it gains from the new features.

Voultapher:
If you want a consistent code style across contributors, I'd **highly** recommend http://editorconfig.org/

SG14/inplace_function.h
```diff
@@ -84,13 +84,27 @@ template<typename R, typename... Args> struct vtable
       ); }
     },
 		copy_ptr{ [](storage_ptr_t dst_ptr, storage_ptr_t src_ptr) -> void
-		  { new (dst_ptr) C{ (*static_cast<C*>(src_ptr)) }; }
+		  {
+        if constexpr (std::is_trivially_copy_constructible<C>::value)
+          std::memcpy(dst_ptr, src_ptr, sizeof(C));
+        else
+          new (dst_ptr) C{ (*static_cast<C*>(src_ptr)) };
+      }
```
Quuxplusone: Does this buy you anything in performance? I would actually expect the compiler to be *better* at inlining a trivial constructor than at inlining a `memcpy`. That is, I would advise removing lines 88,89,90, and 92, unless you can show that they're doing something useful.
Ditto for the move-constructor and destructor cases below: if the `if constexpr` isn't doing something useful (and I will be surprised if it is), I'd recommend to remove it.

SG14/inplace_function.h
```diff
@@ -29,340 +29,331 @@
...
+
+    static_assert(
+      Alignment % std::alignment_of<C>::value == 0,
+      "Incompatible function closure alignment"
+    );
+
+    static const vtable_t vt{detail::wrapper<C>{}};
+    vtable_ptr_ = std::addressof(vt);
+
+    new (std::addressof(storage_)) C{std::forward<C>(closure)};
```
Quuxplusone: Here `std::forward<C>` should be `std::forward<T>`. I wonder if it's possible to write a test case for this?

SG14/inplace_function.h
```diff
@@ -29,340 +29,331 @@
...
+  }
+
+  template<size_t Cap, size_t Align>
+  operator inplace_function<R(Args...), Cap, Align>() &&
+  {
+    static_assert(detail::is_valid_inplace_dst<
+      Cap, Align, Capacity, Alignment
+    >::value);
+
+    return {std::addressof(vtable_ptr_), std::addressof(storage_)};
```
Quuxplusone: The semantic difference between line 301's

    return {std::addressof(vtable_ptr_), std::addressof(storage_)};

and line 291's

    return {vtable_ptr_, std::addressof(storage_)};

is so subtle that I missed it on first reading. I haven't really pondered whether there's a philosophical reason to express "copy" by "copy of the vptr" and "move" by "dereference of the vptr". Might I suggest that you use something like `return inplace_function<R(Args...), Cap, Align>(vtable_ptr_, std::addressof(storage_), std::true_type{})` or something, i.e., overload on a separate tag type to communicate the "should-move-from" bit, instead of overloading on the pointerness of the first argument?

Orthogonally: is there a specific reason that these two are expressed as conversion operators *to* the wider type, as opposed to constructors *from* the narrower type? Personally I rarely use conversion operators, myself, so I'm not necessarily a great judge of when it's appropriate to use them.

SG14/inplace_function.h
carlcook: Good spotting. The typename C is a bit confusing, given that it could be capacity or closure.

carlcook:
A nice implementation! Very interesting that those basic benchmarks give a speedup too. And it's nice now that the boolean conversion operators work.

I'm a little surprised that there's a speedup now that we have moved the invocation pointer into the vtable, as opposed to the original implementation (where the invocation pointer was a first-class member), but maybe this was a micro-optimization that wasn't really needed in the first place.

It might pay to take a look at this snippet: https://github.com/ldionne/dyno/blob/1386582bd63c493ba5aa77d9d07b7b482f35726e/example/function.cpp#L39

This is what Louis Dionne was suggesting after talking to him about inplace function. The library he suggested was this one: https://github.com/ldionne/dyno, but I think this pull request takes a similar approach.

carlcook:
Regarding backwards compatibility for older compilers, I'm in two minds about this. The implementation here is a reference implementation for an upcoming proposal, but is also here to be used if people would like a fairly well tested/complete implementation. For that latter, I agree, some backwards compatibility support would be nice, but I'm not too worried to be honest.

SG14/inplace_function.h
Voultapher: On the conversion construction, it's neither philosophically nor idiomatically backed.
Just found a clearer implementation. When possible I try to avoid parameter type tags.
I've seen them mess with performance due to a changed function signature. Rather express with templates if possible. Plus it puts the type checking on the implementation side, expecting the user of your library to give you the matching tag is brittle.

I opted into conversion function over separate copy ctor, move ctor, copy assignment and move assignment, as it forced repetition or forced shared code. This solution minimizes both.
Also there is the issue of disabling the more specialized generic constructor if it's a `inplace_function` derivative. Which I found doable but ugly if done via the additional constructors and assignments.

SG14/inplace_function.h
Voultapher: I agree.

Voultapher:
Definitely took inspiration from Louis. I'm on bard with him that the standard should adopt a more generic approach of type erased or anything storage related. The abstraction of `storage_policy` he proposes could facilitate a more flexible future library usage. Some of the more common would be inplace, shared and non-owning. This could apply to `function`, `vector` or any other memory constrained library object.

carlcook:
Awesome. Yeah, that's what Louis was getting at with me - i.e. make sure
the proposal can support different storage policies.

Facebook's folly::function is also interesting... it allocates (from what I
can tell) but only supports move-only closures. It would be interesting to
see if this could somehow be supported.

Another comment I've received a few times now is to see if there is a way
of automatically deducing the required closure size, with some form of
decltype wizardry.

SG14/inplace_function.h
Quuxplusone: > expecting the user of your library to give you the matching tag is brittle

Nit: I don't understand this part, since the "user" in this case is just another part of the library class. I wasn't suggesting to expose *public* constructors taking tag types! But whichever; I can understand the rationale about avoiding extra tag-type parameters in signatures, even if I personally would expect that overhead to get inlined away.

On the conversion functions: Ah, makes sense. I hadn't thought about the amount of duplication w.r.t. "copy ctor, move ctor, copy assignment and move assignment."

Voultapher:
@carlcook The capacity and alignment deduction could be done via a user defined deduction guide. Let me implement this tonight.

Regarding the direction of the proposal, I'd be interested in "sitting down" with you and Louis and see if we can come up with a way forward to generalize the current proposal. Following that, we could write a new reference implementation.

Regarding backwards compatiblity, let me see if I can put in some `#ifdef`' around the C++17 only parts.
What name would you suggest for this macro? I guess most people will just copy the file and use it, so putting the language support detection in this build system might not solve the problem.

Quuxplusone:
I think deduction guides for capacity and alignment would be a step backward in terms of usability. I wouldn't want my library *guessing* at the number of bytes it thinks I need, based on seeing a single usage; I'd want it to give me a hard error if I forget to tell it the correct capacity, so that I could quickly go fix it, and not have to debug into hairy compiler messages in order to figure out that in

    int x; long y;
    inplace_function<int()> f = [x](){ return x; };
    inplace_function<int()> g = [y](){ return y; };

`f` and `g` have different types. (Deducing "sizes" of things also smells like John McFarlane's fixed-point literals, in which `10 / 3 == 4`. This could be seen as a pro or as a con.)

Re C++17-only parts, I'd say :shipit: as is (after removing the `if constexpr` parts), and then if there's even any C++17 left, we can easily patch it later to get it compiling as C++14 again. FWIW, the correct `#ifdef` for C++17 is `#if __cplusplus >= 201703L`.

carlcook:
Yeah, I think the conversation about auto-deduction of size came up from
someone in the Bellevue SG14 meeting, but then the discussion quickly
turned to how this could be done, rather than if it was really a good idea
to do it.

carlcook:
@Voultapher Hi Lukas, thanks so much for your work so far.

Regarding user deduced size, if you do have the time to play around with it, let's see how it looks. But @Quuxplusone does raise a good point about this possibly being a hinderance rather than a feature (now that I think about it more).

@nidoizo Have you been following this pull request? What are your thoughts?

@Voultapher I think it's a great idea for us to have a chat with Louis and then maybe rework the reference implementation. I think the key point here is then a really clear second version of the proposal will basically be for free. I'm not too worried about the reference implementation being flawless, but a good reference implementation goes a long way in helping the proposal.

@Quuxplusone Good point about the C++ #ifdefs. Seems a nicer way to go instead of a bunch of custom hashdefs for each supported compiler.

sharpmake:
The new impl will have better perf if the vtable is hot, which might not be the case in a complete application.  I changed the benchmark slightly to use multiple signatures for the functions: http://quick-bench.com/pQErHDfcQLQ3NjQb0U8GkibSN34.

However, I suspect inplace_function without arguments might be frequent, so it's unclear to me what is best.

nidoizo:
The new impl will have better perf if the vtable is hot, which might not be the case in a complete application. I changed the benchmark slightly to use multiple signatures for the functions: http://quick-bench.com/pQErHDfcQLQ3NjQb0U8GkibSN34.

However, I suspect inplace_function without arguments might be frequent, so it's unclear to me what is best.

carlcook:
@nidoizo Interesting benchmark... thanks. But also what do you think about the vtable approach? I quite like the cleaner approach, and am not too worried about the last few nanoseconds of performance either. Are you happy enough to accept this pull request? I am.

Voultapher:
@Quuxplusone thanks for your feedback so far! Something I noticed looking at the libcxx implementation, they seem to [set their version of a `vtable` to 0](https://github.com/llvm-mirror/libcxx/blob/master/include/functional#L1883), aka empty state whenever they are about to copy or move. My best guess is, that its stack unwinding related. Can you come up with a test of a throwing copy assignment, that checks if the cleanup is done right?
My last commit removed all the C++17 bits, with the exception of `inline constexpr`.

Voultapher:
@carlcook @nidoizo Looking at your benchmark, I'm not sure it's vtable hotness related. Afaik [here](http://quick-bench.com/-A6fTxbKB3ntIdhqZBYuEediooA) there is only one `vtable` being used, so it should be hot. Given that I literally never use classical polymorphism, I know too little, to understand all the implications of `TypedCaller` for the benchmark. With the [latest version](http://quick-bench.com/UJIeyqKOq0iJM5KTtx8iPenjLic) of the new implementation the difference doesn't seems as significant.

nidoizo:
@Voultapher I think it's both cache miss and processor prefetch.  If I added more types (16 by adding unsigned combinations) it was making the result difference bigger.  I'm surprised about the single type result, but maybe the additional indirection added is making preprocessor prefetch less likely?

@carlcook I'm fine with implementation as it's not part of the proposal anyway.  I tend to think the old implementation is better performance-wise for a context like the one I have on Rainbow Six Siege, but anyway the most important is avoiding the heap, and for another context the new one could be better.  And as you said the difference is slim.

Voultapher:
@carlcook this review seems stalled, it be great to get some closure or a concrete idea where to go.

carlcook:
@Voultapher I'm happy enough to merge this now. Does anyone else have any thoughts?

I do think that having some form of traits will be useful (as suggested by Louis Dionne), but that is something for a different pull request.

Quuxplusone:
@carlcook: I say merge it. The only person in this thread who's expressed reservations is @nidoizo; @nidoizo do you think we'd be better off by not-merging-this-patch?
If I don't hear objections by 2017-12-08, I'll merge it myself.
(Although, note the merge conflicts. One of us should fix them before merging. And of course when I say "merge" I mean "rebase and squash". ;))

sharpmake:
I'm fine with the merge.  I've nothing more to say than in my previous comment; even if I expect this new impl to be worse performance-wise for some applications, the difference is slim anyway.

nidoizo:
I'm fine with the merge. I've nothing more to say than in my previous comment; even if I expect this new impl to be worse performance-wise for some applications, the difference is slim anyway (sorry for previous post with wrong user).

Voultapher:
...

Quuxplusone:
Sorry, got distracted. :)
I have now merged 3ba4213ebe252519524810646463df4672bee697 et seq!
Thanks @Voultapher!

PR #119, opened 2018-01-12
Improve the struct layout of inplace_function<Sig, 8>.

Quuxplusone:
Attn @carlcook, @Voultapher: this is an ABI-breaking change but I believe it's important to fix.

Before this patch, `inplace_function<Sig, 8>` implicitly meant
`inplace_function<Sig, 8, DefaultAlignment>`, which was
`inplace_function<Sig, 8, 16>` on my MacBook (at least).

After this patch, `inplace_function<Sig, Cap>` implicitly has
the same alignment requirements as `aligned_storage_t<Cap>` (plus
the alignment requirements of the internal vtable pointer type).

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -36,15 +36,6 @@ namespace detail {
 
 static constexpr size_t InplaceFunctionDefaultCapacity = 32;
 
-using default_storage_t = std::aligned_storage_t<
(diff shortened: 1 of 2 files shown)


Voultapher:
@Quuxplusone good catch! Looks right to me.

PR #120, opened 2018-01-19
Add an implementation of Allan Deutsch's slot_map (P0661r2)

Quuxplusone:
Based on the draft paper at
https://gist.github.com/Masstronaut/902319272cd6c5a7b6aa5c5900206f24
with some consultation of Allan's reference implementation at
https://github.com/Masstronaut/slot_array/

Attn: @Masstronaut


Quuxplusone:
https://github.com/Quuxplusone/SG14/compare/slot-map...slot-map-rbegin contains my proposed bonus feature that propagates the "reverse-iterability" of `Container<T>` through `slot_map<T, Key, Container>`.

mattreecebentley:
Would like to see some info (in text file or whatever) to document this, including differences between this and Allan's ver, but otherwise has tacit approval from me - anyone else?

Quuxplusone:
Merged ab9574f910e24a60be28e8148a04cc86ec881e1c, b35c6f1b73f040f877d8cc25c2a74dc4589efe35.

Re documentation on `slot_map` and on things in this repo generally: I think it would be nice to reorganize the directory structure to put each separate "piece" in its own dedicated directory, something like

    include/SG14/slot_map/README.md
    include/SG14/slot_map/slot_map.h
    include/SG14/inplace/README.md
    include/SG14/inplace/inplace_function.h
    tests/

Then there'd be a nice logical place to hang a README for each sub-library. However, I can see how too much structure might turn off some potential contributors; I certainly don't want this to become Boost. :)

Alternatively, we could add a `README-slot_map.md`, `README-inplace_function.md`, etc., either to the root directory here, or perhaps better, in a doc/ directory.

PR #121, opened 2018-01-19
inplace_function: Add tests for functors with throwing copy-constructors.

### LINKED ISSUE #118: inplace_function: unit test for functor copy-assignment that throws
On #116 @Voultapher writes:
> @Quuxplusone thanks for your feedback so far! Something I noticed looking at the libcxx implementation, they seem to [set their version of a vtable to 0](https://github.com/llvm-mirror/libcxx/blob/master/include/functional#L1883), aka empty state, whenever they are about to copy or move. My best guess is, that its stack unwinding related. Can you come up with a test of a throwing copy assignment, that checks if the cleanup is done right?

I have been procrastinating on this. Creating this issue now so that it will not fall through the cracks when #116 is closed/merged.

Quuxplusone:
And with deleted assignment operators, since functor assignment
should never be invoked by `inplace_function`. (Lambdas don't have
assignment operators, for example, and we want to store lambdas.)

Fixes #118.

### DIFF
--- a/SG14_test/inplace_function_test.cpp
+++ b/SG14_test/inplace_function_test.cpp
@@ -260,6 +260,74 @@ void AssignmentDifferentFunctor()
     EXPECT_EQ(4, calls);
 }
 
+struct ThrowingFunctor {
+    static int countdown;
+    static int constructed;
+    static int destructed;
+    static int called;
+    static void reset(int k) {
+        countdown = k;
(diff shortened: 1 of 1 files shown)


Quuxplusone:
Merged as d3514a025da8ef14c027d1809b82a4b5dee06bae

PR #122, opened 2018-01-24
slot_map: Do not require Container to provide reverse_iterator typedefs.

Quuxplusone:
Attn: @masstronaut

With this patch, `slot_map<T,K,C>` will propagate the "reverse-iterability" of `C<T>`, both the existence of the member typedefs and the existence of the member functions `rbegin`, `rend`, `crbegin`, and `crend` (as a group, not individually).

Is this desirable?  I tend to think yes, because implementing reverse-iterability for a custom container is *usually* wasted busywork. OTOH, it could alternatively be addressed by some library (either SG14 or Boost or otherwise) providing a `reversible` mix-in:

    template<class Ctr>
    struct reversible : public Ctr {
        using reverse_iterator = std::reverse_iterator<typename Ctr::iterator>;
        using const_reverse_iterator = std::reverse_iterator<typename Ctr::const_iterator>;
        reverse_iterator rbegin() { return reverse_iterator(this->end()); }
        reverse_iterator rend() { return reverse_iterator(this->begin()); }
        // ... and so on ...
    };

    template<class T> using reVector = reversible<TestContainer::Vector<T>>;
    slot_map<int, std::pair<int,int>, reVector> sm;  // OK

BTW, the fact that this `reVector` alias template can be passed to a template template parameter is *very* cool. I was worried that it wasn't going to be that easy.

### DIFF
--- a/SG14/slot_map.h
+++ b/SG14/slot_map.h
@@ -26,6 +26,7 @@
 
 #pragma once
 
+#include <iterator>
 #include <type_traits>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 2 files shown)


mattreecebentley:
Seems reasonable to me - does anyone have any objections to this? I'm not familiar enough with the code to comment as to side-effects.

PR #123, opened 2018-01-29
slot_map: Proposed improvements to "reserve"/"capacity" API.

Quuxplusone:
Attn: @Masstronaut.

The member function `sm.reserve_slots(n)` used to call `reserve` (and nothing
else) on the slots container. This is not useful functionality. After this
patch, `sm.reserve_slots(n)` resizes the slots container to `n`. It does not
touch the values container, since there is a valid use-case for "many slots,
relatively few values."

The member function `sm.reserve(n)` used to call `reserve` on the values
container. After this patch, `sm.reserve(n)` calls `reserve` on the
values container (and also on the reverse_map container, which is always
the same size as the values container by definition). It *also* resizes
the slots container to `n`, since there is no way for the programmer to
use the `n` capacity of the values container without taking up `n` slots
anyway.

N.B.: For container types that do not provide `reserve`, the methods
`sm.reserve(n)` and `sm.reserve_slots(n)` have exactly the same effects.

The accessor `sm.capacity_slots()` has been renamed to `sm.slot_count()`
for clarity, and returns the `size()` of the slots container.

The accessor `sm.capacity()` still returns the `capacity()` of the
values container.

In the second commit, I make `slot_map` propagate the container's having-of-a-`capacity`-method, so that `sm.capacity()` will exist iff `ctr.capacity()` would be well-formed on the underlying container (and otherwise it will SFINAE away).

### DIFF
--- a/SG14/slot_map.h
+++ b/SG14/slot_map.h
@@ -37,6 +37,28 @@
 
 namespace stdext {
 
+namespace slot_map_detail {
+
+template<size_t I> struct priority_tag : public priority_tag<I-1> {};
+template<> struct priority_tag<0> {};
+
+template<class Ctr, class SizeType>
+inline auto reserve_if_possible(Ctr&, SizeType, priority_tag<0>) -> void {}
+
+template<class Ctr, class SizeType>
+inline auto reserve_if_possible(Ctr& ctr, SizeType n, priority_tag<1>) -> decltype(void(ctr.reserve(n)))
+{
+    ctr.reserve(n);
+}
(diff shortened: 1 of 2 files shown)


Quuxplusone:
Merged a4d4e94ef73eb0418debea8626703a15abaa427f!

WPMGPRoSToTeMa:
@Quuxplusone just interesting. Why do you merge your pull requests by commiting into master and closing PR? But don't via github "merge" button?

PR #129, opened 2018-07-02
Make a moved-from inplace_function actually empty.

### LINKED ISSUE #113: What is the state of a moved-from inplace_function?
Both `folly::Function` and `fu2::function`, as far as I can tell, make the following test case pass:

    function<void()> foo = SomeFunctor{};
    function<void()> bar;
    assert(!bar);
    bar = std::move(foo);  // move the state from foo to bar...
    assert(!foo);  // ...leaving foo empty. (This assertion fails with inplace_function.)

To make this test case fail, you have to leave `foo` in a sort of "partly moved from" state, where calling `foo()` effectively has undefined behavior (because the contained `SomeFunctor` object you're calling is in a moved-from state), which kind of defeats the purpose of having an "empty" (`bad_function_call`-throwing) state at all.

On the other hand, if you *do* make this test case pass, then you end up with a strongly type-safe object type, where the object is fundamentally *not allowed* to enter a "nonsense" or "partly-moved-from" state. This strikes me as a good idea. I think `inplace_function` should follow the example of Folly here, and make this test case pass.

Quuxplusone:
Fixes #113 for real.

(See also: [The Best Type Traits C++ Doesn't Have](https://www.youtube.com/watch?v=MWBfmmg8-Yo). In the vtable lambda for `relocate_ptr`, if we could tell that `C` was _trivially_ relocatable, then we could replace that placement-new-and-pseudo-destructor-call with a single memcpy depending only on `sizeof(C)`, which would then allow us to merge together all the lambdas for a given `sizeof(C)` and reduce the size of the text section.)

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -80,7 +80,7 @@ template<typename R, typename... Args> struct vtable
 
     const invoke_ptr_t invoke_ptr;
     const process_ptr_t copy_ptr;
-    const process_ptr_t move_ptr;
+    const process_ptr_t relocate_ptr;
     const destructor_ptr_t destructor_ptr;
 
     explicit constexpr vtable() noexcept :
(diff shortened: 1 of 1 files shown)


Quuxplusone:
Merged 495a74046d9012f9bfe6551ee48aedf3a79d3ddb!

PR #130, opened 2018-07-28
inplace_function: Use launder to avoid UB

Voultapher:
C++ is 'great' :|

Throwing away the pointer returned by placement new, like we do in `inplace_function` requires us to use `std::launder` to correctly get the memory back, when casting it back to a specific type.

See https://en.cppreference.com/w/cpp/utility/launder for more info.

### DIFF
--- a/SG14/plf_colony.h
+++ b/SG14/plf_colony.h
@@ -1364,7 +1364,7 @@ public:
 				{
 					end_iterator.group_pointer->next_group = PLF_COLONY_ALLOCATE(group_allocator_type, group_allocator_pair, 1, end_iterator.group_pointer);
 					group &next_group = *(end_iterator.group_pointer->next_group);
-					const skipfield_type new_group_size = (total_number_of_elements < static_cast<size_type>(group_allocator_pair.max_elements_per_group)) ? static_cast<const skipfield_type>(total_number_of_elements) : group_allocator_pair.max_elements_per_group;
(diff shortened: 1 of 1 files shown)


Quuxplusone:
I propose that we hold off and do not commit this use of `launder` until we have seen at least one actual STL vendor doing the same thing. I don't think SG14 should be the trailblazer here.

I just verified that libc++'s `std::any` has the same "great" "issue", and does not attempt to use `launder` to fix it.  Ditto libstdc++'s `std::any`, as far as I can tell but I'm not 100% sure. So SG14 is no more broken than any of those folks. And if it ain't broken...

Voultapher:
Fair point, http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0532r0.pdf has some good arguments why `std::launder` might not even solve the problems it set out to solve. Let's keep this PR open, so that other people can see the reasoning for not using `std::launder`. Whenever the 'compiler people' have figured out what they really want, we can either apply or close this PR.

Voultapher:
FWIW, libc++ `map` and `unorder_map` use launder.

Quuxplusone:
> FWIW, libc++ `map` and `unordered_map` use launder.

Yeah, but they're trying to use `launder` to solve the (hypothetical) "problem" caused by `const_cast`. Whereas `any`, `inplace_function`, etc., would be trying to use `launder` to solve the (hypothetical) "problem" caused by `reinterpret_cast`. P0532 implies (in section "Is there more than the constant/reference member issue?") that `launder` was never even *intended* to solve that "problem".

PR #132, opened 2018-09-22
cmake : Update cmake. Apply best-practices.

p-groarke:
There are a few cppcon talks about the new cmake idioms and why they are preferred. I'm happy to share links if you need more convincing ;)

Potentially controversial : Move CMakeLists.txt to root, as is standard in 99.9% of cmake projects.

Other changes:
- Update cmake.
- Use an INTERFACE library for SG14 (header-only library). This allows include directories and options to be inherited when linking with the library (see unit test executable for an example).
- Use modern cmake idioms.
- Use modern compiler checks.
- Don't set CXX_FLAGS that are already set by cmake. Don't use generator time DEBUG/RELEASE. This allows generation of IDE solutions that can easily switch between debug/release.

The changes force C++17 on MSVC, and enable both the new two-phase lookup and the standard c++ version defines. This does cause a rather banal compile issue in colony, and a few warnings (that should be looked at) in slot_map.

I've temporarily disabled warnings as errors on MSVC. If this gets merged, I will open up an issue with Microsoft for the colony problem. I believe it is a VS bug.

Note : I can only test on windows right now. I will be able to test on a mac in the future.


p-groarke:
MSVC bug fix : https://github.com/p-groarke/SG14/commit/8eabdc1c75a8ac3d968a21f27687162fa4a9bb9d

.gitignore
```diff
@@ -71,3 +71,4 @@ cmake/.idea
 /.vs/SG14/v15/Browse.VC.opendb
 /.vs/SG14/v15/Browse.VC.db
 /.vs/ProjectSettings.json
+/[Bb]uild*/
```
Quuxplusone: Could you reasonably restrict this to just `/Build/` and `/build/`? I mean, do you know of a build tool that would create `/build8624/` or `/Build_Foo/` that would need ignoring?

The commit message should explain why `/Build/` and `/build/` need ignoring (i.e., what build tools create them).

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
+		COMPILE_FLAGS "-Wno-unused-parameter"
+	)
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Wfatal-errors -Werror)
+	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
+		COMPILE_FLAGS "-Wno-unused-parameter"
+	)
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
+  # using Intel C++
```
Quuxplusone: Why no warning flags on ICC? Contrariwise, why not just delete lines 70 and 71?

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+	set(CMAKE_LIBRARY_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${BINARY_OUT_DIR})
+	set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${OUTPUTCONFIG} ${BINARY_OUT_DIR})
+endforeach(OUTPUTCONFIG CMAKE_CONFIGURATION_TYPES)
+
+link_directories(${CMAKE_CURRENT_BINARY_DIR}/lib) # For future use with testing library.
+
+##
+# Project
+##
+file(GLOB_RECURSE INCLUDES "${SG14_INCLUDE_DIRECTORY}/*.hpp" "${SG14_INCLUDE_DIRECTORY}/*.h" "${SG14_INCLUDE_DIRECTORY}/*.tpp")
```
Quuxplusone: What is `*.tpp`? I've never seen a project with `.tpp` files. (But I have seen `.hh`, `.H`, and `.ipp`. But not in the SG14 repo, so those extensions shouldn't be in our cmake file either, I wouldn't think.)

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
+cmake_minimum_required(VERSION 3.10)
+project(sg14 CXX)
+
+find_package(Threads REQUIRED)
+
+set(CMAKE_CXX_STANDARD 17)
```
Quuxplusone: I don't suppose there's any way to communicate to CMake that it should use "17 if possible, or 14 otherwise or if specifically requested, or 11 otherwise or if specifically requested..."?  IMHO all of the SG14 stuff should build in '14 and if it doesn't it's probably a bug. (But it should not *break* when built in '17, either, so ideally we'd have a clean way to test both. I don't think CMake is really set up for that kind of thing though, so that's okay.)

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Wfatal-errors -Werror)
+	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
+		COMPILE_FLAGS "-Wno-unused-parameter"
+	)
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
+  # using Intel C++
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT ${TEST_NAME})
+	target_compile_options(${TEST_NAME} PRIVATE /Zc:__cplusplus /permissive- /W4)# /WX) temporarily allow warnings
```
Quuxplusone: > This does cause a rather banal compile issue in colony, and a few warnings (that should be looked at) in slot_map.

Can you copy and paste the error messages into your commit message? That'll help the next person clean them up, and/or just satisfy the reader's curiosity as to what _are_ these mysterious error messages.  Cf. https://github.com/zenorogue/hyperrogue/commit/739146246c5ae1f0a13600bef4b882f9b291465d

CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+	add_definitions(-DNOMINMAX -D_SCL_SECURE_NO_WARNINGS)
+	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
+		COMPILE_FLAGS "/wd4100 /wd4456"
```
Quuxplusone: Likewise it would help to explain in the commit message what are these diagnostic numbers you're suppressing. Cf. https://github.com/zenorogue/hyperrogue/commit/51f59d7bcc919bf8f0a3fbdfc45a228b4a77b6f2

.gitignore
p-groarke: Of course.

CMakeLists.txt
p-groarke: Well, I just haven't ever used the Intel compiler so I don't know the flags. I will remove the lines.

CMakeLists.txt
p-groarke: I will remove it. tpp is often used to separate your template declarations from implementations (I'm guessing it is the same as ipp). At the bottom of your `header.hpp`, you `#include header.tpp`. It is used as a convenience to separate really large headers.

CMakeLists.txt
p-groarke: I don't think either, but I will look into it. I would consider that a "nice to have", maybe in a follow up PR if I get it to work? I will set 14 as the default. I'll probably have to follow up with fixes for MSVC as well.

CMakeLists.txt
p-groarke: Or should I just add the fix? With details in the commit message.

CMakeLists.txt
p-groarke: These are ported from the old cmake file. You would have to ask colony author for the specifics.

p-groarke:
@Quuxplusone Everything should be up to snuff :)

CMakeLists.txt
Quuxplusone: I believe 8d356b7 will fix the MSVC warnings in `slot_map.h`.

The two in `transcode.h` would be easy to fix with a cast, but `transcode.h` is really just a stub, hasn't been touched since its initial commit in October 2016, and I wonder if it's time to just delete it from the master branch. @hatcat may have an opinion.

.gitignore
Quuxplusone: Incidentally, what about the rest of the stuff in the removed `/cmake/.gitignore` — `/cmake/CMakeCache.txt`and such? I infer that the idea is that with a proper build process `CMakeCache.txt` will end up nested under the completely-ignored `/build`, not under `/cmake`, so `CMakeCache.txt` no longer needs to be explicitly ignored, is that right?

I'm skeptical of the ignoring of `/Build`, but maybe that matters on Windows? (But I thought Windows' filesystem might be case-insensitive anyway?)

p-groarke: Correct, everything cmake generated will be in /build, so ignoring that folder is the norm. I've seen people use `Build`, but I don't mind removing it. It isn't a windows thing.

I'm thinking, maybe I should update the readme to indicate the change (`mkdir build && cd build` etc).

CMakeLists.txt
p-groarke: Looks like it. I had planned to fix the warnings after this PR. If there are any left I'll take a look at minimum.

.gitignore
Quuxplusone: > update the readme

Yeah, probably a good idea!  FYI, I always just `cd SG14_test && g++ -std=c++14 -DTEST_MAIN -I../SG14 whatever_test.cpp && ./a.out`; it might be worth mentioning that lazy method in the readme as well.
Describing how to use the cmake file is definitely a good idea, since my understanding starts getting hazy right about where you put the "etc". ;)  Just `mkdir build && cd build && cmake .. && ./test`, something like that?

Quuxplusone:
Btw, do you need someone to commit this for you? I'm inclined to do so; just let me know when it's reached a steady state again.

.gitignore
p-groarke: You can take a look. Nothing exhaustive but its always nice to have copy-paste ready instructions :)

p-groarke:
Yes I would. I'll be testing on a mac soonish. If you are running on linux, would you be willing to test the cmake build?

I'll let you know once its tested.

p-groarke:
If there are no takers by the weekend, I'll setup a VM to test. If you want to test on mac, all you should need is an up-to-date cmake that can find your (relatively) up-to-date compiler :)

p-groarke:
Thx for testing. I pushed the gcc fix + the VS2017 fix. Still a lot of msvc warnings I'd like to resolve.

SG14/plf_colony.h
```diff
@@ -186,7 +186,7 @@
 #include <limits>  // std::numeric_limits
 #include <memory>	// std::allocator
 #include <iterator> // std::bidirectional_iterator_tag
-
+#include <cstddef>
```
WPMGPRoSToTeMa: Maybe add a comment `// offsetof` like for other includes?

p-groarke: Good point, thx

p-groarke:
@Quuxplusone Ready to merge.

WPMGPRoSToTeMa:
@p-groarke sorry, I was incidentally on your master branch... But the error is the same at b2ea5cc (now it is fixed). Also I was need to update my cmake, now I'm on 3.12.2. Now I have the problem with `if constexpr` at the latest commit, because GCC 5.4.0 doesn't support C++17.

p-groarke:
@WPMGPRoSToTeMa I thought colony was checking for constexpr. I will disable the specific warning for VS. I don't want to change the authors container in this PR.

WPMGPRoSToTeMa:
@p-groarke no problems now on GCC 5.4.0.

Issue #131, opened 2018-09-22
[slot_map] partition and underlying_swap


p-groarke:
When using slot_map for components and such, one would usually implement a disable feature. This is done using a pivot, and swapping disabled elements to the end of your contiguous container. Iterating the "enabled" data will thus be contiguous.

I don't think the standard would ever accept a container with this kind of state. So a way to enable this should be provided. From my uses so far, I think a simple `id_swap(Key old, Key new)` would enable a lot of needed flexibility.

For disabling components, one could create 2 `slot_maps`, enabled and disabled, and move objects between the 2. The only problem then is users will have copies of the original id. Adding `id_swap` lets you move into a disabled `slot_map`, swap the id to the old one, never invalidating user owned ids.

There are other uses for `id_swap`. When using RAII to keep user ids valid, the copy constructor would need such a thing.

Ultimately, you want to keep iterators valid when moving/swapping/deleting objects. `id_swap` is just an idea, there may be better ways.

Quuxplusone:
I think the constructor thing should be a separate GitHub issue from the `id_swap` thing.

I doubt I understand your `id_swap` use-case, but superficially it sounds an awful lot like the `std::remove_if` and `std::partition` algorithms which can be used with containers such as `vector`. They could be used with `slot_map` too, if we just add a bit of code to maintain the slot/reversemap indices... so, I've done that (see #133). Would your use-case be adequately addressed by this `partition` member function, or have I misunderstood some aspect of your use-case?

p-groarke:
If I read the partition description correctly, that is the perfect tool for the job. I still believe `id_swap` is essential for RAII, I will come up with a code example shortly.

If you are interested, let me try to clarify expected behavior in a game engine using a `slot_map`-like container. Take for example a class `Baddie`, which is AI enemies. All the baddies are is stored in a `slot_map`. Every frame you call functions (`update`, `late_update`, `physics`, `pre_render`, `render`, `post_render`, etc) on all the baddies. If the `slot_map` is interspersed with disabled baddies, you are getting cache-misses everytime you "jump over" a baddie. So one would partition the `slot_map` first, then call whichever function on only enabled things.

If you are interested in more game engine oriented details, you can read the bitsquid blog series on his entity component system http://bitsquid.blogspot.com/2014/08/building-data-oriented-entity-system.html. From what I read in the `slot_map` paper, it seems the container was in part inspired by those posts.

Quuxplusone:
[...]

PR #133, opened 2018-09-24
Add a slot_map::partition(Pred) member function

Quuxplusone:
Just like the standard `std::partition` algorithm, this member function
reorders the slot_map's elements in such a way that all elements for which
the predicate `pred` returns `true` precede the elements for which `pred`
returns `false`. The relative order of the elements is not preserved.

The function returns an iterator to the first element of the second group
(which may be `end()`, in the case that all elements belong to the first group).

Addresses part of #131, although I'm not sure whether this really addresses
whatever use case @p-groarke is thinking of.


SG14/plf_colony.h
```diff
@@ -1,3867 +1,3867 @@
...
-			if (end_iterator.element_pointer != NULL)
-			{
-				switch(((groups_with_erasures_list_head != NULL) << 1) | (end_iterator.element_pointer == reinterpret_cast<aligned_pointer_type>(end_iterator.group_pointer->skipfield)))
-				{
-					case 0:
-					{
-						const iterator return_iterator = end_iterator;
-
-						#ifdef PLF_COLONY_TYPE_TRAITS_SUPPORT
-							if (std::is_nothrow_move_const
```
WPMGPRoSToTeMa: Line endings was changed (`CR LF` -> `LF`). Did this happen by an accident?

WPMGPRoSToTeMa: Line endings was changed (`CR LF` -> `LF`). Did this happen by an accident?

Quuxplusone: Yeah, `git version 2.7.4 (Apple Git-66)` keeps wanting to change them to match the rest of the repo. [This trainwreck of a StackOverflow question](https://stackoverflow.com/questions/5834014/lf-will-be-replaced-by-crlf-in-git-what-is-that-and-is-it-important) suggests a lot of different fixes, but none of them worked on my system.

    warning: CRLF will be replaced by LF in SG14/plf_colony.h.
    The file will have its original line endings in your working directory.

...Aha! The culprit was that the SG14 repo had a hard-coded `.gitattributes` file checked in! That's not going to work if we've got some people on Windows and some people on Linux. I'm going to WP:BOLD and remove that file.  (It also had some cruft about how to diff .doc files and whatnot, but we don't have any .doc files in the SG14 repo.)  That'll stop git from trying to "fix" @mattreecebentley's line endings all the time.

p-groarke:
TBH, I'm not sure this is need. `std::partition` works great for my use cases.

Quuxplusone:
> TBH, I'm not sure this is need. `std::partition` has worked great for my use cases so far.

That would be great, but off the top of my head, would `std::partition` do the wrong thing because it would break the association between elements and keys?  I mean https://wandbox.org/permlink/NdHnBhZgooDn0raP

    stdext::slot_map<int> sm;
    auto k1 = sm.insert(1);
    auto k2 = sm.insert(2);
    assert(sm.at(k1) == 1);
    assert(sm.at(k2) == 2);
    std::partition(sm.begin(), sm.end(), [](int x) { return x == 2; });
    assert(sm.at(k1) == 2);
    assert(sm.at(k2) == 1);

p-groarke:
Oh yes your right, that's why we needed this. Been a long time :)

SG14/slot_map.h
```diff
@@ -278,6 +278,46 @@ class slot_map
         return 1;
     }
 
+    constexpr void underlying_swap(const_iterator cit, const_iterator cjt) {
```
p-groarke: Comment and complexity would be nice to conform to the rest of the header.

SG14/slot_map.h
```diff
@@ -278,6 +278,46 @@ class slot_map
...
+        auto jt_slot_iter = std::next(slots_.begin(), *jt_reversemap_iter);
+
+        using std::swap;
+        swap(*it, *jt);
+        swap(*it_slot_iter, *jt_slot_iter);
+        swap(*it_reversemap_iter, *jt_reversemap_iter);
+    }
+
+    template<class Pred>
+    constexpr iterator partition(const Pred& pred) {
```
p-groarke: Comments and complexity would be welcome.

p-groarke:
So what was left on this one to be ready for a merge?

p-groarke:
@Quuxplusone Please take a look at https://github.com/WG21-SG14/SG14/issues/147
I discuss some issues I see in `slot_map` which would affect this PR. Cheers

PR #135, opened 2018-09-28
slot_map: The free list should be LIFO, not FIFO

Quuxplusone:
This is mentioned at the end of Allan's paper
http://open-std.org/JTC1/SC22/WG21/docs/papers/2017/p0661r0.pdf

> Alternatively, the free list could store both a front and back index
> on the container, providing the ability to cycle through free slots
> more uniformly, reducing the chance of a generation counter overflow
> occurring. This is a quite valuable benefit with low overhead,
> so explicitly requiring it seems worthwhile.

I'm not 100% sure that this is the best or most efficient (or most correct) way to actually write the code, though.


Quuxplusone:
Merged d47deb39e9e57c5e2220c2ef376b83ea18dcab40!

PR #136, opened 2018-09-29
Fix remainder MSVC warnings

p-groarke:
MSVC warnings that slipped through last PR. Only triggered in debug (due to testing framework).

### DIFF
--- a/SG14/transcode.h
+++ b/SG14/transcode.h
@@ -63,7 +63,7 @@ sg14::lzw_compressor<OutIt, Allocator>::lzw_compressor(OutIt&& dest_, int bits_)
 template<typename OutIt, typename Allocator>
 sg14::lzw_compressor<OutIt, Allocator>& sg14::lzw_compressor<OutIt, Allocator>::operator=(const char& element)
 {
-	OutIt::operator=(element + 1);
+	OutIt::operator=(static_cast<char>(element + 1));
 	return *this;
 }
 
@@ -75,7 +75,7 @@ sg14::lzw_decompressor<OutIt, Allocator>::lzw_decompressor(OutIt&& dest_)
 template<typename OutIt, typename Allocator>
 sg14::lzw_decompressor<OutIt, Allocator>& sg14::lzw_decompressor<OutIt, Allocator>::operator=(const char& element)
 {
-	OutIt::operator=(element - 1);
+	OutIt::operator=(static_cast<char>(element - 1));
 	return *this;
 }
 
@@ -89,7 +89,7 @@ sg14::rsa_encryptor<OutIt>::rsa_encryptor(OutIt&& dest_, int m_, int n_)
 template<typename OutIt>
 sg14::rsa_encryptor<OutIt>& sg14::rsa_encryptor<OutIt>::operator=(const char& element)
 {
-	OutIt::operator=(element + m);
+	OutIt::operator=(static_cast<char>(element + m));
 	return *this;
 }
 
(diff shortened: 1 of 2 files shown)


SG14/transcode.h
```diff
@@ -89,7 +89,7 @@ sg14::rsa_encryptor<OutIt>::rsa_encryptor(OutIt&& dest_, int m_, int n_)
 template<typename OutIt>
 sg14::rsa_encryptor<OutIt>& sg14::rsa_encryptor<OutIt>::operator=(const char& element)
 {
-	OutIt::operator=(element + m);
+	OutIt::operator=(static_cast<const char>(element + m));
```
WPMGPRoSToTeMa: Can it be just `char` instead of `const char`?

p-groarke: sure

WPMGPRoSToTeMa:
Hm, why it doesn't warn here?
https://github.com/WG21-SG14/SG14/blob/203a411f4863e3724df15c8f08c2d35cd672f618/SG14/transcode.h#L66

p-groarke:
@WPMGPRoSToTeMa I'm not sure. It could be that it isn't covered by unit tests (thus never triggered), or maybe the constant is implicitly converted to char?

WPMGPRoSToTeMa:
@p-groarke even `char + char` is `int` due to integral promotion. Looks like it is a MSVC bug. Can you try GCC with `-Wconversion`?

p-groarke:
I don’t have gcc (windows only here). However, I am working on another PR for CI (osx, Linux, and windows). If that is the case, then the issue will be caught in CI.

Are you OK to put this on the backburner till we get CI in?

p-groarke:
Oh I didn't see your edit. Will fix those as well then :)

WPMGPRoSToTeMa:
> Are you OK to put this on the backburner till we get CI in?

There are also enough errors in `plf_colony.h` with `-Wconversion` enabled, but I'm ok with fixing them later.

SG14_test/ring_test.cpp
```diff
@@ -175,19 +175,19 @@ static void reverse_iterator_test()
     r.push_back(2);
     r.push_back(3);
     r.push_back(4);
-    std::vector<int> v(3);
+    std::vector<double> v(3);
```
Quuxplusone: I think this change is incorrect, or at least, significant. The old code tested that `std::copy` from `ring<double>` to `vector<int>` worked; the new code tests only from `ring<double>` to `vector<double>` (i.e., between iterators sharing the same `value_type`).

I think the compiler is being overzealous here. However, would it be satisfactory (to both you and the compiler) to switch the types so that we're copying from a `ring<int>` into a `vector<double>`, or maybe from `ring<char>` into `vector<int>`?

SG14/transcode.h
```diff
@@ -103,6 +103,7 @@ sg14::rsa_decryptor<OutIt>::rsa_decryptor(OutIt&& dest_, int m_, int n_)
 template<typename OutIt>
 sg14::rsa_decryptor<OutIt>& sg14::rsa_decryptor<OutIt>::operator=(const char& element)
 {
-	OutIt::operator=(element - m);
+	OutIt::operator=(static_cast<char>(element - m));
```
Quuxplusone: I'll merge this diff right now, even though I would still like @hatcat to think about moving this entire header into a non-master branch and/or deleting it.

p-groarke: thx, just trying to get it to build on vs2017

p-groarke:
@Quuxplusone Hey, did you see the updates?

Quuxplusone:
Merged 871017b53c7ae7b9de2b713e3c1d20a9e6d0015a!

PR #137, opened 2018-10-06
CI

p-groarke:
Enables CI and testing with gcc 5, 6, 7, 8 (linux), clang 4.0.x (macos), msvc 2017 (win). Both debug and release builds are tested.

To whomever may merge : You will need to add the repo on travis and appveyor. Once that is done, the readme should be updated with your badge links as well.

New commits or PRs will trigger CI automatically.

Some changes of note:
- The threaded `ring` test has been modified so it doesn't rely on user input.
- NDEBUG is undefined to allow unit testing in release mode.
- Some `slot_map` warnings were disabled for gcc 7. It freaks out with structured bindings (gcc 8 will catch real c++17 warnings).
- Previous `plf_colony` fixes were reverted. I believe the author ignored them when pushing a new update. They have been reapplied. CI would have caught this ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,11 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
 	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
 		COMPILE_FLAGS "-Wno-unused-parameter"
 	)
+	if (CMAKE_CXX_COMPILER_VERSION MATCHES "^7.*")
+		set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/slot_map_test.cpp PROPERTIES
+			COMPILE_FLAGS "-Wno-error=unused-variable -Wno-error=unused-but-set-variable") # Fix gcc7 issues with structured bindings
+		message("Disabled -Wunused-variable and -Wunused-but-set-variable for gcc ${CMAKE_CXX_COMPILER_VERSION}.")
+	endif()
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
 	set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT ${TEST_NAME})
 	target_compile_options(${TEST_NAME} PRIVATE /Zc:__cplusplus /permissive- /W4 /WX)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,12 +61,9 @@ target_include_directories(${TEST_NAME} PRIVATE "${SG14_TEST_SOURCE_DIRECTORY}")
 
 # Compile options
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
-	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Wfatal-errors -Werror)
-	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
-		COMPILE_FLAGS "-Wno-unused-parameter"
-	)
+	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Werror)
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Wfatal-errors -Werror)
+	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Werror)
 	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
 		COMPILE_FLAGS "-Wno-unused-parameter"
 	)
@@ -80,6 +77,5 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
 	target_compile_options(${TEST_NAME} PRIVATE /Zc:__cplusplus /permissive- /W4 /WX)
 	add_definitions(-DNOMINMAX -D_SCL_SECURE_NO_WARNINGS)
 	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
-		COMPILE_FLAGS "/wd4100 /wd4456 /wd4127"
-	)
+		COMPILE_FLAGS "/wd4127") # Disable conditional expression is constant, use if constexpr
 endif()
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
 # SG14
+[![Build Status](https://travis-ci.org/WG21-SG14/SG14.svg?branch=master)](https://travis-ci.org/WG21-SG14/SG14)
+
 A library for Study Group 14 of Working Group 21 (C++)
 
 /docs - Documentation for implementations without proposals, or supplementary documentation which does not easily fit within a proposal.
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,133 @@
+language: cpp
+
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      brew update;
+      brew upgrade cmake;
+      brew install cmake;
+    fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+      eval "${MATRIX_EVAL}";
+      DEPS_DIR="${TRAVIS_BUILD_DIR}/deps";
+      mkdir ${DEPS_DIR} && cd ${DEPS_DIR};
+      CMAKE_URL="https://cmake.org/files/v3.10/cmake-3.10.0-Linux-x86_64.tar.gz";
+      mkdir cmake && travis_retry wget --no-check-certificate --quiet -O - ${CMAKE_URL} | tar --strip-components=1 -xz -C cmake;
+      export PATH=${DEPS_DIR}/cmake/bin:${PATH};
+      cd ..;
+    fi
+
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode9.2
+      env:
+        - CONFIG=Release
+
+    - os: osx
+      osx_image: xcode9.2
+      env:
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -23,4 +26,4 @@ Clone the repo. Navigate to the folder in your favorite terminal.
 `cmake .. && cmake --build . && ./bin/sg14_tests`
 
 ### Alternatively
-`cd SG14_test && g++ -std=c++14 -DTEST_MAIN -I../SG14 whatever_test.cpp && ./a.out`
\ No newline at end of file
+`cd SG14_test && g++ -std=c++14 -DTEST_MAIN -I../SG14 whatever_test.cpp && ./a.out`
```
WPMGPRoSToTeMa: New line was added?

CMakeLists.txt
```diff
@@ -61,20 +61,21 @@ target_include_directories(${TEST_NAME} PRIVATE "${SG14_TEST_SOURCE_DIRECTORY}")
...
+	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Werror)
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Wfatal-errors -Werror)
+	target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Werror)
 	set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
 		COMPILE_FLAGS "-Wno-unused-parameter"
 	)
+	if (CMAKE_CXX_COMPILER_VERSION MATCHES "^7.*")
+		set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/slot_map_test.cpp PROPERTIES
+			COMPILE_FLAGS "-Wno-error=unused-variable -Wno-error=unused-but-set-variable") # Fix gcc7 issues with structured bindings
```
WPMGPRoSToTeMa: Can `[[maybe_unused]]` solve the problem?

mattreecebentley:
Compiles fine on msvc2017 here with warning level 4 - I don't follow these CR's much because the main repo is github/mattreecebentley for colony etc. Have commented on your change for that and will update with next release.
As for authors reverting changes - yes, that happens - often people have things in more than one repo, and things get out of step - it's nothing personal.

I'm not familiar with CI, but it seems you should point out where and when the current repo isn't compiling.

Quuxplusone:
I had been meaning to reply but it fell off my radar. I want to express vague support for the idea of CI. In specific, I think the only problem with adding CI would be @mattreecebentley's code — he's previously expressed the idea that at-least-parts-of this repo should be more experimental playgrounds and not necessarily "forced" to adhere to arbitrary and artificial requirements such as "passing the unit tests" or "building." ;)

> I'm not familiar with CI, but it seems you should point out where and when the current repo isn't compiling.

Well, that's basically the idea of CI. The CI provider (e.g. TravisCI) would automatically download and build the code periodically, and put a red badge on the GitHub page if it didn't build. However, that approach works only if someone (e.g. @mattreecebentley) periodically *checks* the color of the badge and acts on it. If checking and acting is seen as drudge work, then it probably just won't get done, and the badge will remain red forever.

Personally, I think that if SG14's repo isn't the source of truth for `plf_colony`, then `plf_colony` shouldn't be distributed via SG14's repo. It should be a git submodule, or just a link in the README. Right now SG14 is distributing a "slightly out-of-date mirror" of `plf_colony`, which doesn't seem like useful behavior to me (personally).

johnmcfarlane:
@p-groarke I've added you to WG21-SG14 as a member in case that helps you set up the necessary hooks.

p-groarke:
Would having a branch experimental be adequate for experimentation? I think you can disable CI per branch. Another alternative to submodules are cmake's ExternalProject, which clone a repo at build time. They work much better than submodules for external dependencies imho. Finally, one could simply add experimental headers to an experimental subfolder and ignore those in testing/building.

Let me know what the consensus is, if it is worth persuing the CI effort or not. Thanks a lot for the prompt response.

@mattreecebentley The issue with the if statement was problematic as MSVC seemed to be having problems with less-than/greater-than and parsing them as templates. Maybe this is fixed in the latest VS release, I will update. If we go ahead with CI, and it is still a problem, I'll send you a link to the CI error.

mattreecebentley:
Arthur, stop being creepy.
Anyway, I don't have a problem with it *in principle*, so long as it doesn't create any additional work for contributors - specifically new contributors. That way lies boost. Arthur is misrepresenting my position.

The problem with a CI is that it assumes the compilers it's testing are the only compilers. It's much more useful if people actually tell devs where and on what compiler the code is failing - then it comes down to whether it's a compiler bug (which has happened) or dev work required.

Making an experimental branch is a dumb idea. Whatevers in the repo should be publishable code. I don't put code up that doesn't compile in the 5 compilers I test with.

SG14_test/ring_test.cpp
```diff
@@ -61,36 +63,43 @@ static void thread_communication_test()
...
 		int val = 0;
+		size_t check_pos = 0;
 		do
 		{
 			std::unique_lock<std::mutex> lk(m);
-			cv.wait(lk);
+			cv.wait(lk, [&]() { return !buffer.empty(); }); // spurious wakeups
 			val = buffer.front();
-			std::cout << val << std::endl;
+			assert(val == values[check_pos++]);
```
Quuxplusone: Presumably this `check_pos++` should be outside the `assert`.

p-groarke:
> Making an experimental branch is a dumb idea. Whatevers in the repo should be publishable code. I don't put code up that doesn't compile in the 5 compilers I test with.

Not only is that not enough, that's a waste of your time. CI can do that automatically for you.

mattreecebentley:
No, it doesn't. You're presuming what you'd be doing is a superset of what I'm doing. In no way is that the case. Nor should it be.

Quuxplusone:
I just merged appropriate changes to `slot_map.h` and `ring_test.cpp`, so you can remove those files from this PR.

I'm inclined to say let's go ahead and merge this PR. As you point out, it'll need someone to go fiddle with Travis and Appveyor. I just now used Travis to trigger a request to the owners of @WG21-SG14 asking them to enable Travis access. I cannot grant that access myself.

carlcook:
I received a link to grant just then (and have approved it).

Cheers,
Carl

mattreecebentley:
I'm not sure that it'll help me, but it might help some others.
Regardless, there needs to be one person responsible for deleting/updating compilers when obvious compilers bugs like the one mentioned are found. Otherwise it's a no-go, from my perspective.
Also, the one non-compiler bug (as of C++20) noted for colony has been fixed and merged as of last week- 
Cheers

Quuxplusone:
> Regardless, there needs to be one person responsible for deleting/updating compilers when obvious compiler bugs like the one mentioned are found. Otherwise it's a no-go, from my perspective.

Yes, one problem with running tests is that someone has to fix them when they go red.

I will say, though, that one advantage of running tests is that it gives a quantifiable line between "the code works" and "the code doesn't work." There's no wiggling out of a red test by saying "oh, that's not _my_ bug, it's a _compiler_ bug." If the tests are red, someone has to get them green again — even if that means writing code tailored to work even on "buggy" compilers. (But sometimes, in practice, what it means is removing that entry from the test matrix.)

@p-groarke: Can you split out the TravisCI part (not including the AppVeyor part) into a new PR, and make it as close to merge-ready as you can? (Including any fixes to the badge URLs, for example.) Then I'll merge it at least long enough to see if the badge is green.

p-groarke:
@Quuxplusone Yep, I'll get everything ready soonTM. I'll ping you when it's ready to merge. Splitting the travis is going to be a hassle imo, are you ok if I don't display the appveyor badge for now, but leave the config in?

Quuxplusone:
> Splitting the travis is going to be a hassle imo, are you ok if I don't display the appveyor badge for now, but leave the config in?

@p-groarke: Why would it be a hassle? Wouldn't it just be `git rm appveyor.yml` and removing the badge? FWIW I am no CMake expert; is it the case that some of the CMake changes are motivated by AppVeyor support and would be poorly motivated without AppVeyor?

p-groarke:
@Quuxplusone It's more about git and the rebasing hell that it'll incure. I'll squash the PR, then it should just be `git rm appveyor.yml` as you mention.

mattreecebentley:
> I will say, though, that one advantage of running tests is that it gives a quantifiable line between "the code works" and "the code doesn't work." There's no wiggling out of a red test by saying "oh, that's not _my_ bug, it's a _compiler_ bug." If the tests are red, someone has to get them green again — even if that means writing code tailored to work even on "buggy" compilers. (But sometimes, in practice, what it means is removing that entry from the test matrix.)
> 

In practice, we've already seen with p-groarke merge request that there're extremely OBVIOUS compiler bugs coming up, that should not be worked around, but reported. MS and clang are usually pretty responsive on this (https://github.com/mattreecebentley/plf_colony/issues/15).
No, someone needs to put their hand up to be the 'go to' person when this fails.

p-groarke:
Maybe its none of my business to pitch in, but in the industry at large CI is considered just like unit tests or bugs. If you break it you fix it. Usually it is run on PRs, so you don't have a choice to fix issues before merging ;)

p-groarke:
@Quuxplusone Everything should be up to snuff (removed duplicate work, removed appveyor, updated badge). Here is are the results of latest CI run on this PR https://travis-ci.org/p-groarke/SG14

I'll open the MSVC pr once this is merged.

Quuxplusone:
@carlcook: It looks like TravisCI isn't building anything (when it should be building the "master" branch and pull requests). I wonder if your earlier authorization timed out after a few days without a `.travis.yml`, and if re-authorizing would fix it. Can you take a look?
https://travis-ci.org/WG21-SG14/SG14

TBBle:
I just 'Activated' this repository on the [travis-ci.org side](https://travis-ci.org/WG21-SG14/SG14) and triggered a [custom build from master](https://travis-ci.org/WG21-SG14/SG14/builds/462605138). I'm not sure if the previous setup expired while waiting for a `travis.yml`, or a button-click was missed during setup.

Either way, it *should* now react to pull-requests and pushes to master (the only branch right now).

Quuxplusone:
:tada: Looks like the build passed! [And the badge is green.](https://github.com/WG21-SG14/SG14#sg14) So now it's just a matter of making sure nobody turns it red (or if they do, persuading them to turn it back).

PR #138, opened 2018-11-06
Make inplace_function's move operations "noexcept"

### LINKED ISSUE #128: inplace_function move constructor should be noexcept
Right now, the following program performs poorly:
https://wandbox.org/permlink/vLEl9Owmp1exrkKA

    struct Widget {
        Widget() {}
        Widget(Widget&&) noexcept { puts("move"); }
        Widget(const Widget&) noexcept { puts("copy"); }
    };

    auto f = [w = Widget()]() {};
    using IPF = stdext::inplace_function<void(), 32>;
    std::vector<IPF> v;
    v.emplace_back(f);
    v.emplace_back(f);
    [...]

The vector resize does `copy` instead of `move`, because `inplace_function`'s move constructor is not marked `noexcept`, and vector is pessimized in that case.

I'd like to see us explicitly drop support for non-noexcept-moveable types, preferably by specifying that the move constructor will call `std::terminate` if the wrapped object throws during its move-construction. Then the patch to de-pessimize the vector code above would be a single keyword: just add `noexcept` to the move constructor. (And to `swap`; and optionally to move-assignment.)

@carlcook thoughts?

Quuxplusone:
Fixes #128.

The new test case fails before, and passes after, this change to `inplace_function.h`.

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -104,7 +104,7 @@ template<typename R, typename... Args> struct vtable
             { ::new (dst_ptr) C{ (*static_cast<C*>(src_ptr)) }; }
         },
         relocate_ptr{ [](storage_ptr_t dst_ptr, storage_ptr_t src_ptr)
-            noexcept(std::is_nothrow_move_constructible<C>::value) -> void
+            noexcept -> void
             {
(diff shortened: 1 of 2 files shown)


Quuxplusone:
Merged 5c875e394938c68248ff8e771de6cb8cc768702c!

PR #142, opened 2018-12-12
slot_map : Const find doesn't contain out of bounds check.

p-groarke:
@Quuxplusone Would probably want to use Scott Meyer's const_cast trick in the future?


Quuxplusone:
Merged 2a756bff57cc9d1a2983e94d3e316dd3fe01397d with a new unit test that fails before the patch and passes after. Thanks!

PR #143, opened 2019-01-05
slot_map : DRY, container access and cleanup.

p-groarke:
@Quuxplusone 
To prevent issues like the const find out-of-range access that happen recently, use dry best practices. I didn't use the const_cast technique for one-liner getters. Happy to add that if you want it.

Make container access public, it is required for C interop.

### DIFF
--- a/SG14/slot_map.h
+++ b/SG14/slot_map.h
@@ -114,11 +114,7 @@ public:
     // O(1) time and space complexity.
     //
     constexpr reference at(const key_type& key) {
-        auto value_iter = this->find(key);
-        if (value_iter == this->end()) {
-            SLOT_MAP_THROW_EXCEPTION(std::out_of_range, "at");
-        }
-        return *value_iter;
+        return const_cast<reference>(static_cast<const slot_map*>(this)->at(key));
     }
     constexpr const_reference at(const key_type& key) const {
         auto value_iter = this->find(key);
@@ -132,24 +128,20 @@ public:
     // If the check fails it is undefined behavior.
     // O(1) time and space complexity.
     //
-    constexpr reference operator[](const key_type& key)              { return *find_unchecked(key); }
-    constexpr const_reference operator[](const key_type& key) const  { return *find_unchecked(key); }
+    constexpr reference operator[](const key_type& key) {
+        return const_cast<reference>(static_cast<const slot_map*>(this)->operator [](key));
+    }
+    constexpr const_reference operator[](const key_type& key) const {
+        return *find_unchecked(key);
+    }
 
     // The find() functions have generation counter checking.
     // If the check fails, the result of end() is returned.
     // O(1) time and space complexity.
     //
     constexpr iterator find(const key_type& key) {
-        auto slot_index = get_index(key);
-        if (slot_index >= slots_.size()) {
-            return end();
-        }
-        auto slot_iter = std::next(slots_.begin(), slot_index);
-        if (get_generation(*slot_iter) != get_generation(key)) {
-            return end();
-        }
-        auto value_iter = std::next(values_.begin(), get_index(*slot_iter));
-        return value_iter;
+        const_iterator it = static_cast<const slot_map*>(this)->find(key);
+        return values_.erase(it, it);
     }
     constexpr const_iterator find(const key_type& key) const {
         auto slot_index = get_index(key);
@@ -168,9 +160,8 @@ public:
     // O(1) time and space complexity.
     //
     constexpr iterator find_unchecked(const key_type& key) {
-        auto slot_iter = std::next(slots_.begin(), get_index(key));
-        auto value_iter = std::next(values_.begin(), get_index(*slot_iter));
-        return value_iter;
+        const_iterator it = static_cast<const slot_map*>(this)->find_unchecked(key);
+        return values_.erase(it, it);
     }
     constexpr const_iterator find_unchecked(const key_type& key) const {
         auto slot_iter = std::next(slots_.begin(), get_index(key));
--- a/SG14/slot_map.h
+++ b/SG14/slot_map.h
@@ -295,13 +295,12 @@ public:
         swap(next_available_slot_index_, rhs.next_available_slot_index_);
     }
 
-protected:
     // These accessors are not part of P0661R2 but are "modernized" versions
     // of the protected interface of std::priority_queue, std::stack, etc.
-    constexpr Container<mapped_type>& c() & noexcept { return values_; }
-    constexpr const Container<mapped_type>& c() const& noexcept { return values_; }
-    constexpr Container<mapped_type>&& c() && noexcept { return std::move(values_); }
-    constexpr const Container<mapped_type>&& c() const&& noexcept { return std::move(values_); }
+    constexpr Container<mapped_type>& container() & noexcept { return values_; }
+    constexpr const Container<mapped_type>& container() const& noexcept { return values_; }
+    constexpr Container<mapped_type>&& container() && noexcept { return std::move(values_); }
+    constexpr const Container<mapped_type>&& container() const&& noexcept { return std::move(values_); }
 
 private:
     constexpr slot_iterator slot_iter_from_value_iter(const_iterator value_iter) {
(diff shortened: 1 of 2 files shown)


SG14/slot_map.h
```diff
@@ -114,11 +114,7 @@ class slot_map
     // O(1) time and space complexity.
     //
     constexpr reference at(const key_type& key) {
-        auto value_iter = this->find(key);
-        if (value_iter == this->end()) {
-            SLOT_MAP_THROW_EXCEPTION(std::out_of_range, "at");
-        }
-        return *value_iter;
+        return const_cast<reference>(static_cast<const slot_map*>(this)->at(key));
```
Quuxplusone: The built-in `const_cast` is not guaranteed to be able to cast between `const_reference` and `reference`, unless they happen to be native references — which might be essentially required anyway (I think it is, for allocator types, at least?) but I'd rather avoid `const_cast` anyway.

SG14/slot_map.h
```diff
@@ -132,24 +128,20 @@ class slot_map
     // If the check fails it is undefined behavior.
     // O(1) time and space complexity.
     //
-    constexpr reference operator[](const key_type& key)              { return *find_unchecked(key); }
-    constexpr const_reference operator[](const key_type& key) const  { return *find_unchecked(key); }
+    constexpr reference operator[](const key_type& key) {
+        return const_cast<reference>(static_cast<const slot_map*>(this)->operator [](key));
```
Quuxplusone: This seems to replace a simple expression with a crazy complicated one. I hope on second glance you'd agree that the old code was actually more DRY? ;)
Irrelevant nitpick: in general I'd write `(*this)[key]` where you wrote `this->operator[](key)`.

SG14/slot_map.h
```diff
@@ -132,24 +128,20 @@ class slot_map
...
-            return end();
-        }
-        auto slot_iter = std::next(slots_.begin(), slot_index);
-        if (get_generation(*slot_iter) != get_generation(key)) {
-            return end();
-        }
-        auto value_iter = std::next(values_.begin(), get_index(*slot_iter));
-        return value_iter;
+        const_iterator it = static_cast<const slot_map*>(this)->find(key);
+        return values_.erase(it, it);
```
Quuxplusone: This diff seems wrong.
If this diff is indeed wrong, _and_ if the unit-tests don't currently catch it, would you mind opening a new PR with a single new unit test that would have caught this?

SG14/slot_map.h
```diff
@@ -304,13 +295,12 @@ class slot_map
...
     // These accessors are not part of P0661R2 but are "modernized" versions
     // of the protected interface of std::priority_queue, std::stack, etc.
-    constexpr Container<mapped_type>& c() & noexcept { return values_; }
-    constexpr const Container<mapped_type>& c() const& noexcept { return values_; }
-    constexpr Container<mapped_type>&& c() && noexcept { return std::move(values_); }
-    constexpr const Container<mapped_type>&& c() const&& noexcept { return std::move(values_); }
+    constexpr Container<mapped_type>& container() & noexcept { return values_; }
+    constexpr const Container<mapped_type>& container() const& noexcept { return values_; }
+    constexpr Container<mapped_type>&& container() && noexcept { return std::move(values_); }
+    constexpr const Container<mapped_type>&& container() const&& noexcept { return std::move(values_); }
```
Quuxplusone: > Make container access public, it is required for C interop.

This diff seems fine, but I wish there were a bit more explicit rationale given, for posterity. What is a realistic example of something you want to do but can't, with the `protected` interface?

(Full disclosure: I believe nothing in a well-designed class should ever be `protected`. Public seems like the right way to go here, a priori. But I was also trying to remain close-ish to the existing container adaptors, which have a protected _data_ member named `c`. Also, giving completely uninhibited access to the underlying container, such that any client code can `insert` into it arbitrarily, seems a bit _too_ wild-west to me. I wish there were some slightly more scary/verbose way of getting the container than the harmless-looking `sm.container()`.)

Suppose we removed these four overloads and replaced them with a single pilfering overload:

    constexpr Container<mapped_type> container() && noexcept { return std::move(values_); }

Would that meet your requirements in re "C interop"?

SG14_test/slot_map_test.cpp
```diff
@@ -86,6 +86,24 @@ struct Vector {
...
+            return data_beg;
+        }
+        iterator data_end = data_beg + delta;
+        iterator it = data_beg;
+        while (it != data_end) {
+            it->~T();
+            ++it;
+        }
+        iterator new_end = end() - delta;
+        std::swap_ranges(data_beg, data_end, new_end);
```
Quuxplusone: This seems like it would call `swap` on a destroyed object.

SG14_test/slot_map_test.cpp
```diff
@@ -86,6 +86,24 @@ struct Vector {
         a = std::move(b);
         b = std::move(t);
     }
+    iterator erase(const_iterator first, const_iterator last) {
```
Quuxplusone: Why is this member function required? (If it's really required, add a unit test that fails before this addition and succeeds afterward.)

You should also update [the exposition-only SlotMapContainer concept](https://github.com/WG21-SG14/SG14/blob/f53bc87714cc3f18049a98d15ddae8cdadec867a/SG14_test/slot_map_test.cpp#L541) if `erase` is required.

SG14/slot_map.h
p-groarke: The `const_cast` is required for this to work. It is one of the rare cases where `const_cast` is safe. Are you thinking of an alternative?

SG14/slot_map.h
p-groarke: Oops. Will remove this one. I agree it is extreme for one liners :)

SG14/slot_map.h
p-groarke: What do you mean by "the diff is wrong"? Erase is used to convert from const_iterator to iterator only here. It is the recommended technique to "const_cast" an iterator.

SG14/slot_map.h
p-groarke: This has popped up a few times while working with `slot_map`. One good example is pushing data to your gpu in opengl. Which expects a pointer + size. I would be OK with a `data` function, but that seems just as dangerous as `container` in my mind.

I don't think you should force the user to inherit the container for something as simple as reading the data. It is true it's very "wild-west", but isn't that what we love? "Here is a shotgun, go play with it." ;)

Your suggestion does not work, as you do not want to move the data out, just access it. Maybe we could start with a const version first and see if that fulfills most needs? The code that requires this is at work, so I'll check if a const only version would work and get back to you on that.

SG14_test/slot_map_test.cpp
p-groarke: It is to convert a `const_iterator` to `iterator`. I'll add it to the concept test.

SG14_test/slot_map_test.cpp
p-groarke: Good point, will fix that up. Thx!

SG14/slot_map.h
Quuxplusone: Oh, I see, you're erasing a range of length _zero_ (not length one). Still, rather than require the container to provide a "`const_iterator_cast`" spelled `erase`, we should either not require `const_iterator_cast` at all (my preference), or else require it to be provided via a specific customization point.

SG14/slot_map.h
Quuxplusone: If you just want a `data, size` pair, and you know the underlying container is contiguous, then what's wrong with `&sm.front(), sm.size()`?

Alternatively, I think it would be reasonable to expose a `.data()` method (conditionally, iff the underlying container has a `.data()` method).

But it sounds like for your purposes `sm.empty() ? nullptr : &sm.front()` would be 100% fine as a "pointer to the data."

SG14/slot_map.h
p-groarke: huh, I had no idea `vector<bool>` references behaved that way. I don't think a solution to that particular problem exists. I'll close the PR then as there is nothing else to salvage.

SG14/slot_map.h
p-groarke: Fair point, I don't remember if there was another case, but that should indeed work.

Issue #141, opened 2018-12-11
[slot_map] Can you get a key using an iterator?


p-groarke:
I'm trying to find the keys of stored objects that match a given predicate. I can't see any api to get a key from an iterator, but I may be missing something.

Is it possible to get a key using an iterator in `slot_map`?

p-groarke:
OK I believe that is provided by `slot_iter_from_value_iter`, it is simply marked private.

Quuxplusone:
This, like #131, sounds like a legitimate design issue. @Masstronaut, are you still interested in `slot_map`? Are these features good ideas, or did you consider them and then exclude them from the `slot_map` proposal for reasons that it would be worthwhile to recap here?

@p-groarke: If the functionality you think-you-need is there-but-private, then this issue should still be open, right? Either we should make the functionality public, or we should explain why you don't really need it, or we should explain why if you do need it you shouldn't be using `slot_map`.

My kneejerk reaction is that we should make it public, which means bikeshedding the name/API. Like,

    slot_map<K, V> sm;
    std::vector<V>::iterator it = sm.begin();
    K k = sm.key(it);  // new functionality -- is this a good API?
    assert(sm.find(k) == it);

p-groarke:
@Quuxplusone Fair point, until we get an answer from the author and windows CI, my hands are a little tied. I have many features I'd like to add. I'm happy to send a PR for this one if the original author is on board / doesn't have objections.

Quuxplusone:
@p-groarke: If you have "many features" you'd like to add, I'd recommend forking the repo, adding all the features in a branch, and then submitting the whole thing as a big patch so we can see how all the new things work together. Bugfixes = incremental, design work = holistic. :)  (#135 is certainly related, if we're doing design work.)

FWIW, I haven't seen any comments on this repo from @Masstronaut (Allan Deutsch) for a long time. I myself am the original author of this repo's "slot_map.h", which I did just by slavishly following Allan's [P0661R0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0661r0.pdf).  Allan has his own `slot_map` at https://github.com/Masstronaut/slot_array .

I see there is also https://github.com/twiggler/slotmap . I don't know anything about it except that it has a *lot* of heavy dependencies: `std::random_device`, Boost.Integer, Boost.Hana,...

p-groarke:
[...]

PR #145, opened 2019-01-18
Slot map - Add unit tests, conform operator[]() to comment, add at_unchecked(), add key()

p-groarke:
- Adds a few missing unit tests.

- `operator[]` didn't conform to the comment. Moved completely unchecked behavior to `at_unchecked` and modified `operator[]` to match comment. It has a generation check, but no out-of-bounds check.

- As mentioned in https://github.com/WG21-SG14/SG14/issues/141, add a key api to find the key of a given element.

- Renamed `key_size_type` to `key_index_type`.

### DIFF
--- a/SG14/slot_map.h
+++ b/SG14/slot_map.h
@@ -128,17 +128,6 @@ public:
         return *value_iter;
     }
 
-    // The at_unchecked() function has no checks whatsoever
-    // and does not throw.
-    // O(1) time and space complexity.
-    //
-    constexpr reference at_unchecked(const key_type& key) {
-        return *find_unchecked(key);
-    }
-    constexpr const_reference at_unchecked(const key_type& key) const {
-        return *find_unchecked(key);
-    }
-
     // The bracket operator[] has a generation counter check
     // and does not throw.
     // If the check fails it is undefined behavior.
--- a/SG14/slot_map.h
+++ b/SG14/slot_map.h
@@ -128,12 +128,38 @@ public:
         return *value_iter;
     }
 
-    // The bracket operator[] has a generation counter check.
+    // The at_unchecked() function has no checks whatsoever
+    // and does not throw.
+    // O(1) time and space complexity.
+    //
+    constexpr reference at_unchecked(const key_type& key) {
+        return *find_unchecked(key);
+    }
+    constexpr const_reference at_unchecked(const key_type& key) const {
+        return *find_unchecked(key);
+    }
+
+    // The bracket operator[] has a generation counter check
+    // and does not throw.
     // If the check fails it is undefined behavior.
     // O(1) time and space complexity.
     //
-    constexpr reference operator[](const key_type& key)              { return *find_unchecked(key); }
-    constexpr const_reference operator[](const key_type& key) const  { return *find_unchecked(key); }
+    constexpr reference operator[](const key_type& key) {
+        auto slot_iter = std::next(slots_.begin(), get_index(key));
+        if (get_generation(*slot_iter) != get_generation(key)) {
+            return *end();
+        }
+        auto value_iter = std::next(values_.begin(), get_index(*slot_iter));
+        return *value_iter;
+    }
+    constexpr const_reference operator[](const key_type& key) const {
+        auto slot_iter = std::next(slots_.begin(), get_index(key));
+        if (get_generation(*slot_iter) != get_generation(key)) {
+            return *end();
+        }
+        auto value_iter = std::next(values_.begin(), get_index(*slot_iter));
+        return *value_iter;
+    }
 
     // The find() functions have generation counter checking.
     // If the check fails, the result of end() is returned.
@@ -302,6 +328,16 @@ public:
         swap(values_, rhs.values_);
         swap(reverse_map_, rhs.reverse_map_);
         swap(next_available_slot_index_, rhs.next_available_slot_index_);
+    }
+
+    // O(1) time and space complexity.
+    //
+    constexpr key_type key(const_iterator value_iter) const {
+        auto value_index = std::distance(const_iterator(values_.begin()), value_iter);
+        auto slot_index = *std::next(reverse_map_.begin(), value_index);
+        auto slot_it = std::next(slots_.begin(), slot_index);
+        auto generation = get_generation(*slot_it);
+        return key_type{ slot_index, generation };
     }
 
 protected:
--- a/SG14_test/slot_map_test.cpp
+++ b/SG14_test/slot_map_test.cpp
@@ -1,26 +1,36 @@
 #include "SG14_test.h"
 #include "slot_map.h"
-#include <assert.h>
-#include <inttypes.h>
 #include <algorithm>
+#include <assert.h>
 #include <deque>
 #include <forward_list>
-#include <list>
+#include <inttypes.h>
 #include <iterator>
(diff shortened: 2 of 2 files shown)


SG14/slot_map.h
```diff
@@ -128,12 +128,38 @@ class slot_map
...
+    constexpr reference at_unchecked(const key_type& key) {
+        return *find_unchecked(key);
+    }
+    constexpr const_reference at_unchecked(const key_type& key) const {
+        return *find_unchecked(key);
+    }
+
+    // The bracket operator[] has a generation counter check
+    // and does not throw.
     // If the check fails it is undefined behavior.
```
Quuxplusone: The original comment may be unnecessarily confusing, but I believe the original comment and code are correct, here. "`operator[]` has a check, and if that check fails, you get UB" is basically the same thing as "`operator[]` has no check."

(By analogy to `vector`, the proper name for the unchecked/UB-prone version of `at()` _is_ `operator[]`.)

SG14_test/slot_map_test.cpp
```diff
@@ -1,26 +1,36 @@
...
 
 namespace TestKey {
 struct key_16_8_t {
     uint16_t index;
     uint8_t generation;
+    friend bool operator==(key_16_8_t lhs, key_16_8_t rhs);
 };
+bool operator==(key_16_8_t lhs, key_16_8_t rhs) {
+    return lhs.index == rhs.index && lhs.generation == rhs.generation;
+}
```
Quuxplusone: Please move this friend function in-line (use the "hidden friend idiom"), and also please change its name to something other than `operator==`, just so that it's clear that it's not part of the expected API of a key type. I suggest `friend bool KeysAreEqual(key_16_8_t lhs, key_16_8_t rhs)`.

If you do that, and pull the `slot_map::key(const_iterator)` member function into its own PR (or equivalently, remove the `at_unchecked` part from this PR), I'm favorably disposed toward it.

SG14/slot_map.h
p-groarke: IMHO it isn't the same, as your debug build will catch the undefined behavior with debug iterators. Just as accessing vector out-of-range is caught in debug.

SG14_test/slot_map_test.cpp
p-groarke: I'll remove `at_unchecked`, do you also want me to remove the `operator[]` changes? I'm quite opposed to that. The current subscript implementation is quite dangerous.

p-groarke: Great blog post about ADL, is this a ploy to drive traffic to your website? lol!

When looking up the hidden friend idiom, I found this : http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1601r0.pdf

So the idiom just states to make the operator inline friend. If I understand your ADL post correctly, this limits the exposure of the operator to ADL only.

I am curious what is the problem case that this fixes, when the operator is exposed in the enclosing namespace (declared out-of-line)?

p-groarke:
@Quuxplusone I updated the PR with your comments. I removed `at_unchecked`, I left the changes for `operator[]`. I think asserting would be ok with the old behavior, but as you mentioned before, adding a dependency to the `cassert` header isn't ideal. I don't think the subscript operator should happily and quietly return the wrong element, without a way to notify the user. The current implementation should fail in debug, without requiring asserts, and mimics accessing a vector out-of-range.

I couldn't use `KeysAreEqual` as the test runs on `std::pair` keys. So I added a comment to clarify + used hidden friends. Should be a good compromise.

SG14/slot_map.h
Quuxplusone: Hmm. I had not been thinking about the observable behavior re debug iterators. That's a thought-provoking wrinkle.

So in master right now there are two ways to index into a slot_map: `sm.at(key)` (throws on key-not-found) and `sm[key]` (UB on key-not-found).  And there are two ways to index into a vector: `vec.at(idx)` (throws on out-of-range) and `vec[idx]` (UB on out-of-range, except on MSVC with `_ITERATOR_DEBUG_LEVEL != 0`, where it calls `_STL_REPORT_ERROR("vector subscript out of range")`).

So I think you're asking for `sm[key]` to go via a codepath that, when `_ITERATOR_DEBUG_LEVEL != 0`, will reliably trigger `_STL_REPORT_ERROR("something-or-other")` on key-not-found. Which your PR does implement, because `*end()` will trigger `_STL_REPORT_ERROR`.

However, I don't want regular users of `sm[key]` (outside of debug mode) to pay for the check on your line 138:

        if (get_generation(*slot_iter) != get_generation(key)) {
            return *end();
        }

I guess if you just enclosed those three lines in `#if _ITERATOR_DEBUG_LEVEL != 0`, then that would accomplish your goal. Feels icky to me, though.

p-groarke: Yeah I find that icky as well. I think this is worth hashing out, because it speaks to the fundamental understanding of `slot_map` and its behavior.

For me, `slot_map` is a map. If I give a bad key to a map, I expect things to "break". Returning an unrelated object (the same one every time), isn't undefined behavior. It is quite predictable, you get the last object that was at that slot. I guess you could argue that any behavior is undefined behavior, but that is a stretch ;)

Lets say you completely ignore the generation here, then the map accepts non-existant keys and happily returns the object at that slot. Doesn't this completely redefine what a `slot_map` is supposed to be?

Performance-wise, the `slot_map` excels at iterating multiple objects in the underlying container. I don't believe a single `if` is problematic, especially if you compare to a hash + collision you might get from `unordered_map<size_t, T>`. Which is what a `slot_map` replaces.

Quuxplusone: > Returning an unrelated object (the same one every time), isn't undefined behavior. It is quite predictable, you get the last object that was at that slot.

That is, you get a reference to a destroyed object, or a reference to an unrelated object (if the slot has been reused), or a segfault (if the underlying vector has been reallocated and the memory not reused for a different object yet), or a plain old wild pointer (if the memory has been reused). Similarly, with `std::vector::operator[]`, when you access an invalidated iterator, you get a reference to a destroyed object, or a reference to an unrelated object (if the slot has been reused), or a segfault or a wild pointer.  All of these possibilities are conforming manifestations of "undefined behavior."

In MSVC's debug mode, you pay a hefty runtime cost to detect invalid accesses and make them produce predictable `assert`-like behavior. In release mode, all the cost goes away.

I wouldn't strongly object to adding such a "debug mode" to `slot_map`. But I _would_ want to make sure that in "release mode," all the cost goes away.

> Lets say you completely ignore the generation here, then the map accepts non-existant keys and happily returns the object at that slot. Doesn't this completely redefine what a slot_map is supposed to be?

Indexing into a core-language array with `a[10000]` happily returns the object `10000*sizeof(a[0])` bytes beyond the start of `a`. Does that "completely redefine" what an array is supposed to be?  I say "no." Undefined behavior does sometimes do the "expected" thing. But it has the freedom to do the "wrong" thing, too, and we exploit that freedom to gain a little extra runtime performance.

p-groarke: OK fair enough. I'll remove that part of the PR. Will squash as well.

I guess I should ask one last time, would you be willing to add an `assert` here? I know you didn't want to add a dependency on `<cassert>`, but I feel like it is fully justifiable here.

Issue #149, opened 2019-02-03
[inplace_function] Ambiguous overloads.


p-groarke:
Just hit an issue when overloading a function accepting `inplace_function` vs. `std::function`. The compiler (MSVC 2017, clang 7) cannot disambiguate by function arguments. Here is a repro example and a link to it on compiler explorer.

Not sure if there is something to be done about this, worth investigating though.

```C++
#include <cstdio>
#include <functional>
#include <sg14/inplace_function.h>

// Something to be done about this?

/**
 * Ambiguous
 */
void myfunc(stdext::inplace_function<void()>&& exec) {
    std::invoke(exec);
}
void myfunc(stdext::inplace_function<void(size_t)>&& exec) {
    std::invoke(exec, size_t(42));
}

/**
 * Not ambiguous
 */
void myfunc_std(std::function<void()>&& exec) {
    std::invoke(exec);
}
void myfunc_std(std::function<void(size_t)>&& exec) {
    std::invoke(exec, size_t(42));
}

int main(int, char**) {
    // ambiguous
    myfunc([](){
        printf("blee\n");
    });

    // ok
    myfunc_std([](){
        printf("blou\n");
    });

    return 0;
}
```

https://godbolt.org/z/38cCZz

Good day

Voultapher:
I've given a whole talk about why using `std::function` as function parameter is a terrible idea, the same applies to `stdext::inplace_function`. The only viable use case for them is for storing runtime dynamic layout objects, usually as class member variables. There SFINAE does not apply, so I'd say we should not encourage bad usage.

As function parameters, invocable objects should be templates. Among other benefits, this allows the user to introspection himself, even more powerful than what would be possible with overload sets.
Example:
```cpp
class SomeWorkQueue
{
public:
  template<typename T> auto push_back(T func) -> void
  {
    if constexpr (sizeof(T) <= 4)
    {
      _small_queue.push_back(func);
    }
    else if constexpr (sizeof(T) <= 8)
    {
      _medium_queue.push_back(func);
    }
    else
    {
      // error handling etc
    }
  }
private:
  std::vector<stdext::inplace_function<void(), 4>> _small_queue;
  std::vector<stdext::inplace_function<void(), 8>> _medium_queue;
};
```

p-groarke:
@Voultapher That's great and all, but you don't give any compelling reason why. I personally use templates where it makes sense.

I guess people who have to support a stable ABI and update user libraries should just choose another language? Or change job?

[...]

Issue #125, opened 2018-02-14
inplace_function implicit conversion chooses copy over move


Quuxplusone:
This is a really unfortunate side-effect of a core language issue, IMNSHO; this week I have been actively working on trying to understand and diagnose the core issue. [CWG 1579](http://wg21.link/cwg1579) is related but its resolution left much to be desired.

Consider this program:

    using IPF20 = stdext::inplace_function<void(), 20>;
    using IPF40 = stdext::inplace_function<void(), 40>;
    static_assert(std::is_convertible<IPF20, IPF40>::value, "");

    IPF40 foo() {
        IPF20 f;
        return f;  // HERE
    }

What happens on the line marked "HERE"?  Naively, we would expect NRVO (copy elision). But that can't happen because IPF20 and IPF40 are not the same type. So, post-CWG1579, we would expect lvalue-to-rvalue conversion — we'd expect that the expression `f` is treated as an rvalue, and the wrapped functor is *moved* from `f` into the return slot. However, CWG1579 kicks in only for *constructors*; it does not kick in for `inplace_function`'s clever use of a [templated conversion operator](https://github.com/WG21-SG14/SG14/blob/98baf1a/SG14/inplace_function.h#L294-L312) to implement implicit conversions. So CWG1579 does *not* kick in, and what we actually get in this case is a *copy operation*.

SG14 could fix this by changing the implementation of implicit conversions for `inplace_function` to use converting constructors instead of conversion operators, but this strikes me as a library fix for a language bug. I am pessimistic but not fatalistic that we might actually be able to get a core language fix here. I'm opening this GitHub issue just to document the issue for posterity.

(I remember remarking on @Voultapher's unusual and clever use of conversion operators at the time. He convinced me that it was the Right Way to do it, and I still think it is, really. But I'm sad to see that it actually *does* have minor negative consequences due to this core language bug.)

mattreecebentley:
Thanks for making me look up IMNSHO :)
Assumption of copy elision never works out well, in my experience. While it'd be nice to be able to rely on compilers and the language always doing the right thing, it's better to brute-force it if you always want the right outcome.

PR #155, opened 2019-03-31
Make inplace_function SFINAE aware

Voultapher:
@Quuxplusone @p-groarke this should address the issues described in #149.

I went with a custom implementation of `std::is_invocable_r`, because the standard versions don't seem to respect return types, and there seem to be pre C++17 users of this library.

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -150,6 +150,39 @@ struct is_valid_inplace_dst : std::true_type
     );
 };
 
+// C++11 MSVC compatible implementation of std::is_invocable_r.
+
+template<class R> void accept(R);
+
+template<class, class R, class F, class... Args> struct is_invocable_r_impl : std::false_type {};
+
+template<class F, class... Args> struct is_invocable_r_impl<
+    decltype(std::declval<F>()(std::declval<Args>()...), void()),
+    void,
+    F,
+    Args...
+> : std::true_type {};
+
+template<class F, class... Args> struct is_invocable_r_impl<
+    decltype(std::declval<F>()(std::declval<Args>()...), void()),
+    const void,
+    F,
+    Args...
+> : std::true_type {};
+
+template<class R, class F, class... Args> struct is_invocable_r_impl<
+    decltype(accept<R>(std::declval<F>()(std::declval<Args>()...))),
+    R,
+    F,
+    Args...
+> : std::true_type {};
+
+template<class R, class F, class... Args> using is_invocable_r = is_invocable_r_impl<
+    void,
+    R,
+    F,
+    Args...
+>;
 } // namespace inplace_function_detail
 
 template<
@@ -190,15 +223,13 @@ public:
     template<
         typename T,
         typename C = std::decay_t<T>,
-        typename = std::enable_if_t<!inplace_function_detail::is_inplace_function<C>::value>
+        typename = std::enable_if_t<
+            !inplace_function_detail::is_inplace_function<C>::value
+            && inplace_function_detail::is_invocable_r<R, C, Args...>::value
+        >
     >
     inplace_function(T&& closure)
     {
-#if __cplusplus >= 201703L
-        static_assert(std::is_invocable_r<R, C, Args...>::value,
-            "inplace_function cannot be constructed from non-callable type"
-        );
-#endif
         static_assert(std::is_copy_constructible<C>::value,
             "inplace_function cannot be constructed from non-copyable type"
         );
--- a/SG14_test/inplace_function_test.cpp
+++ b/SG14_test/inplace_function_test.cpp
@@ -29,7 +29,7 @@ struct ConstFunctor {
     void operator()(int i) const { assert(i == expected); called_with = i; }
 };
 
-void Foo(int i)
+static void Foo(int i)
 {
     assert(i == expected);
     called_with = i;
@@ -41,14 +41,14 @@ std::string gLastS;
 int gLastI = 0;
 double gNextReturn = 0.0;
 
-double GlobalFunction(const std::string& s, int i)
+static double GlobalFunction(const std::string& s, int i)
 {
     gLastS = s;
     gLastI = i;
     return gNextReturn;
 }
 
-void FunctionPointer()
+static void FunctionPointer()
 {
     // Even compatible function pointers require an appropriate amount of "storage".
     using CompatibleFunctionType = std::remove_reference_t<decltype(GlobalFunction)>;
@@ -65,7 +65,7 @@ void FunctionPointer()
     EXPECT_EQ(42, gLastI);
 }
 
-void Lambda()
+static void Lambda()
 {
     stdext::inplace_function<double(int), 8> fun;
     std::string closure("some closure");
@@ -80,7 +80,7 @@ void Lambda()
     EXPECT_EQ(42, gLastI);
 }
 
-void Bind()
+static void Bind()
 {
     stdext::inplace_function<double(int), 64> fun;
     std::string closure("some closure");
@@ -111,7 +111,7 @@ struct AnotherFunctor
 int AnotherFunctor::mDestructorCalls = 0;
 int AnotherFunctor::mConstructorCalls = 0;
 
-void FunctorDestruction()
+static void FunctorDestruction()
 {
     AnotherFunctor::mDestructorCalls = 0;
     AnotherFunctor::mConstructorCalls = 0;
@@ -140,7 +140,7 @@ void FunctorDestruction()
     EXPECT_EQ(AnotherFunctor::mDestructorCalls, AnotherFunctor::mConstructorCalls);
 }
 
-void Swapping()
+static void Swapping()
 {
     AnotherFunctor::mDestructorCalls = 0;
     AnotherFunctor::mConstructorCalls = 0;
@@ -166,7 +166,7 @@ void Swapping()
     EXPECT_EQ(AnotherFunctor::mDestructorCalls, AnotherFunctor::mConstructorCalls);
 }
 
-void Copying()
+static void Copying()
 {
     auto sptr = std::make_shared<int>(42);
     EXPECT_EQ(1, sptr.use_count());
@@ -190,7 +190,7 @@ void Copying()
     EXPECT_TRUE(bool(fun2));
 }
 
-void ContainingStdFunction()
+static void ContainingStdFunction()
 {
     // build a big closure, bigger than 32 bytes
     uint64_t offset1 = 1234;
@@ -210,7 +210,7 @@ void ContainingStdFunction()
     EXPECT_EQ(r, int(offset1+offset2+offset3+str1.length()+3));
 }
 
-void SimilarTypeCopy()
+static void SimilarTypeCopy()
 {
     auto sptr = std::make_shared<int>(42);
     EXPECT_EQ(1, sptr.use_count());
@@ -236,7 +236,7 @@ void SimilarTypeCopy()
     fun4 = fun1; // fun1 is bigger than 17, but we should be smart about it
 }
 
-void AssignmentDifferentFunctor()
+static void AssignmentDifferentFunctor()
 {
     int calls = 0;
     stdext::inplace_function<int(int,int), 16> add = [&calls] (int a, int b) { ++calls; return a+b; };
@@ -395,7 +395,7 @@ static void test_overloaded_operator_new()
     EXPECT_EQ(43, fun(1));
 }
 
-void test_move_construction_is_noexcept()
+static void test_move_construction_is_noexcept()
 {
     using IPF = stdext::inplace_function<void(int), sizeof(Functor)>;
     std::vector<IPF> vec;
@@ -407,7 +407,7 @@ void test_move_construction_is_noexcept()
     EXPECT_EQ(1, moved);
 }
 
-void test_move_construction_from_smaller_buffer_is_noexcept()
+static void test_move_construction_from_smaller_buffer_is_noexcept()
 {
     using IPF32 = stdext::inplace_function<void(int), 32>;
     using IPF40 = stdext::inplace_function<void(int), 40>;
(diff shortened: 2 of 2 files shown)


Voultapher:
I force pushed some indentations fixes, can someone re trigger the CI.

Quuxplusone:
> I force pushed some indentations fixes, can someone re trigger the CI.

(Force-pushed to your feature branch, I assume and hope. ;)) My understanding is that GitHub will notice and re-trigger the Travis CI build on its own, relatively quickly, any time the tip of the feature branch changes. Whether the history "changed" or just "incremented" doesn't matter to GitHub.

SG14/inplace_function.h
```diff
@@ -150,6 +150,76 @@ struct is_valid_inplace_dst : std::true_type
     );
 };
 
+// C++11 MSVC compatible implementation of std::is_invocable_r,
+// std version doesn't seem to respect return types and requires C++17.
+// See https://godbolt.org/z/T0bd3K for defect example.
```
Quuxplusone: By "doesn't seem to respect return types ... defect example," IIUC, you mean this?
https://godbolt.org/z/YMT0PA

    static_assert( std::is_invocable_r_v<void, int()> );  // passes

That is, although `int` is not "convertible to" `void`, nonetheless `int()` is still "invocable with return type convertible to" `void`. This makes *some* intuitive sense, since you can always discard the result of a function call; and in fact this is what `std::function` does [[1]](https://quuxplusone.github.io/blog/2019/03/27/design-space-for-std-function/#can-i-implicitly-convert-f-int-t) [[2]](https://github.com/zenorogue/hyperrogue/commit/f6f96e1129df134a9921015dc6e59d45ce5a7e11). Some people probably want this behavior, so that you can write things like

    std::function<void(const char*)> f = std::puts;

But *I'm* an anecdotal example of someone who *wouldn't* want such a conversion to happen implicitly.

A priori, I'd recommend we follow `std::function`'s example and permit conversions such as

    stdext::inplace_function<void(const char*)> f =
        [](const char *) { return 0; };

even though it will make some people (such as myself) unhappy.

SG14/inplace_function.h
```diff
@@ -150,6 +150,76 @@ struct is_valid_inplace_dst : std::true_type
...
+static_assert(! is_invocable_r<int, C_Int2>::value, "IV B");
+static_assert(! is_invocable_r<int, C_Void>::value, "IV C");
+
+static_assert(is_invocable_r<int, C_Int2, int>::value, "IV D");
+static_assert(! is_invocable_r<int, C_Int1, int>::value, "IV E");
+static_assert(! is_invocable_r<int, C_Void, int>::value, "IV F");
+
+static_assert(is_invocable_r<void, C_Void, int&>::value, "IV G");
+static_assert(! is_invocable_r<void, C_Int1, int&>::value, "IV H");
+static_assert(! is_invocable_r<void, C_Int2, int&>::value, "IV I");
```
Quuxplusone: I would be inclined to move all these test cases out of the header file and into the unit-tests. I know I set a bad example already with `aligned_storage`, though. 😛 

If they're kept here, at least I think they should be simplified down to just in-line function types, i.e.

    static_assert(is_invocable_r<int, int()>::value, "IV A");
    static_assert(!is_invocable_r<int, int(int)>::value, "IV B");
    static_assert(!is_invocable_r<int, void(int&)>::value, "IV C");

That's easier to read, and doesn't introduce new entities into the user's program, and I don't think it loses anything important in terms of coverage.

SG14_test/inplace_function_test.cpp
```diff
@@ -429,6 +429,48 @@ static void RvalueRefParameter()
...
+    return out_parameter;
+}
+
+int fobau(NoDefaultCtor) { return 2; }
+
+
+void test_overloading()
+{
+    using brokenType = decltype(fobau(std::declval<NoDefaultCtor>()));
+    EXPECT_EQ(brokenType{2}, 2);
```
Quuxplusone: I don't understand the point of this test case at all. `fobau` isn't overloaded, so it's super obvious that `decltype(fobau(std::declval<NoDefaultCtor>()))` is `int`, and of course `int{2} == 2`. I suspect testing/scaffolding code that was meant to be removed.

SG14_test/inplace_function_test.cpp
```diff
@@ -429,6 +429,48 @@ static void RvalueRefParameter()
...
+
+    EXPECT_EQ(overloaded_function([]() -> int { return 3; }), 3);
+    EXPECT_EQ(overloaded_function([](int arg) -> int { return arg; }), 42);
+    EXPECT_EQ(overloaded_function([](int arg, NoDefaultCtor ndc) -> int { return arg + ndc.val; }), 42 + 5);
+    EXPECT_EQ(overloaded_function([](int& arg) -> void { arg += 3; }), 88 + 3);
```
Quuxplusone: Personally, I would rather see these test cases as simple type-trait tests:

    auto a = []() -> int { return 3; };
    assert(( std::is_convertible<decltype(a), stdext::inplace_function<int()>>::value ));
    assert(( !std::is_convertible<decltype(a), stdext::inplace_function<int(int)>>::value ));
    assert(( !std::is_convertible<decltype(a), stdext::inplace_function<void(int&)>>::value  ));
    auto b = [](int&) -> void {};
    assert(( !std::is_convertible<decltype(b), stdext::inplace_function<int()>>::value ));
    assert(( !std::is_convertible<decltype(b), stdext::inplace_function<int(int)>>::value ));
    assert(( std::is_convertible<decltype(b), stdext::inplace_function<void(int&)>>::value ));

SG14/inplace_function.h
```diff
@@ -150,6 +150,76 @@ struct is_valid_inplace_dst : std::true_type
...
+
+template<
+    typename R,
+    typename C,
+    typename... Args
+>
+using def_closure_return_type_convertible = std::is_convertible<
+    typename std::decay<R>::type,
+    decltype(std::declval<C>()(std::declval<Args>()...))
+>;
```
Quuxplusone: This feels overly complicated. What if it were just something like (UNTESTED CODE):
```
template<class R> void accept(R);

template<class, class R, class F, class... Args> struct is_invocable_r_impl : std::false_type {};
template<class F, class... Args> struct is_invocable_r_impl<decltype(declval<F>()(declval<Args>()...), void()), void, F, Args...> : std::true_type {};
template<class F, class... Args> struct is_invocable_r_impl<decltype(declval<F>()(declval<Args>()...), void()), const void, F, Args...> : std::true_type {};
template<class R, class F, class... Args> struct is_invocable_r_impl<decltype(accept<R>(declval<F>()(declval<Args>()...))), R, F, Args...> : std::true_type {};

template<class R, class F, class... Args> using is_invocable_r = is_invocable_r_impl<void, R, F, Args...>;
```

SG14_test/inplace_function_test.cpp
Voultapher: My bad, that's leftover code, I'll remove it.

SG14/inplace_function.h
Voultapher: Sigh, another day goes by and C++'s unreasonable compromises win once again.

SG14/inplace_function.h
Voultapher: My intuition was to put a test in test file, yet I didn't want to break with the precedent `aligned_storage` set so I put them there, I'll gladly move them to the test file.

SG14_test/inplace_function_test.cpp
Voultapher: Ok, are you sure that tests the exact same thing, as in can that cause an ambiguous overload error? Also why `assert` and not `static_assert`?

SG14/inplace_function.h
Voultapher: Whether that's simpler I guess depends on your tmp style, this avoids the use of `std::is_convertible` which fails with incomplete types on mac. So I'll try this.

SG14_test/inplace_function_test.cpp
Quuxplusone: Tests the same thing: Yes, in that if only exactly one of the `is_convertible`s is true, then of course the overload resolution won't be ambiguous. However, neither your original tests nor mine would handle (perhaps impossible-to-implement) things where both conversions are *possible* but have different overload-resolution priorities anyway, such as

    A foo(inplace_function<int()>);
    B foo(inplace_function<long()>);
    int bar();
    A a = foo(bar);  // test that IPF<int()> is a better match than IPF<long()>
    // std::function does not support this, and I think it might be impossible to support

    A foo(inplace_function<void(const int&)>);
    B foo(inplace_function<void(int&)>);
    void bar(const int&);
    A a = foo(bar);  // test that IPF<void(const int&)> is a better match than IPF<void(int&)>
    // std::function does not support this, and I think it might be impossible to support



`assert` instead of `static_assert`: brain fart. 😛

Quuxplusone: https://cplusplus.github.io/LWG/issue2813 is related.

Voultapher:
Ok, that should address your last comment regarding the conversion tests. I went with a hybrid approach, that should catch most problems, and added some additional closure type tests.

Voultapher:
Friendly ping, is there something blocking this?

SG14_test/inplace_function_test.cpp
```diff
@@ -429,6 +429,77 @@ static void RvalueRefParameter()
...
+};
+
+int overloaded_function(stdext::inplace_function<int()>&& exec)
+{
+    return exec();
+}
+
+int overloaded_function(stdext::inplace_function<int(int)>&& exec)
+{
+    return exec(int{42});
```
Quuxplusone: Just `exec(42)`, please. (Right?)

SG14_test/inplace_function_test.cpp
```diff
@@ -429,6 +429,77 @@ static void RvalueRefParameter()
...
+
+using C_Int1 = int();
+using C_Int2 = int(int);
+using C_Void = void(int&);
+
+using stdext::inplace_function_detail::is_invocable_r;
+
+static_assert(is_invocable_r<int, C_Int1>::value, "IV A");
+static_assert(! is_invocable_r<int, C_Int2>::value, "IV B");
+static_assert(! is_invocable_r<int, C_Void>::value, "IV C");
```
Quuxplusone: I'd like to see additional checks somewhere in here for

    static_assert(is_invocable_r<void, C_Int1>::value);
    static_assert(is_invocable_r<long, C_Int1>::value);
    static_assert(is_invocable_r<char, C_Int1>::value);

(testing widening and narrowing conversions, and the "conversion" to `void`). Also,

    static_assert(! is_invocable_r<int, C_Void, int&>::value);

(testing the conversion *from* `void` to `int`, which should definitely not be allowed).

Finally, I think we should have regression tests for both `is_invocable` and `is_convertible`:

    static_assert(is_invocable_r<const int&, int()>::value);
    static_assert(is_convertible<int(*)(), inplace_function<const int&()>>::value);

[These tests do currently pass for `std::function`](https://godbolt.org/z/OP23Ae) but my understanding is that at least the second line is expected to fail after the adoption of whatever paper is resolving [LWG 2813](https://cplusplus.github.io/LWG/issue2813).

Voultapher: `static_assert(is_convertible<int(*)(), inplace_function<const int&()>>::value, "E2")` fails, and I don't quite understand why.

Quuxplusone: The culprit is the `std::is_convertible<C, inplace_function>::value` here:

    template<
        typename T,
        typename C = std::decay_t<T>,
        typename = std::enable_if_t<
            !(std::is_same<C, inplace_function>::value
            || std::is_convertible<C, inplace_function>::value)
            && inplace_function_detail::is_invocable_r<R, C, Args...>::value
        >
    >
    inplace_function(T&& closure)

This is already problematic in master:

    static_assert(std::is_convertible<int(*)(), stdext::inplace_function<int()>>::value, "");

The problem is that you're using `is_convertible<C, ipf>` as a SFINAE condition in order to decide whether to _make_ `C` convertible to `ipf`. I think this is only working by accident. Anyway, in this specific case, `is_convertible` needs to decide whether there's any implicit constructor `inplace_function(int(*arg)())` (it knows there's no implicit conversion operator from `int(*)()` to `inplace_function`). So it does template substitution into this SFINAE condition, which causes the instantiation of `std::is_convertible<C, inplace_function> [with C=int(*)()]`, which makes a cycle in the instantiation graph. I don't know if this even has defined behavior; but in practice what it does is make `std::is_convertible<C, inplace_function>` come out to `false_type`. And then that `false_type` gets memoized by the compiler and re-used for the condition of the `static_assert` later on. Does that make sense?

I think the solution is going to have to be to turn the conversion operators back into constructors. Which we already want to do because of #125.

Voultapher:
It feels like this pr is spiraling out of scope. Given that the remaining problems exist on master and are somewhat unrelated to the changes in this pr, I'd suggest merging this as is and addressing #125 in a separate pr.

SG14_test/inplace_function_test.cpp
Quuxplusone: @Voultapher wrote:
> It feels like this pr is spiraling out of scope. Given that the remaining problems exist on master and are somewhat unrelated to the changes in this pr, I'd suggest merging this as is and addressing #125 in a separate pr.

I agree that this PR is spiraling out of scope. However, I strongly recommend that we fix the known issues with master (i.e. fix `std::is_convertible<int(*)(), stdext::inplace_function<int()>>`) *before* introducing new and compounded issues (i.e. enabling people to SFINAE on that already-wrong expression).

That means we should fix `std::is_convertible<int(*)(), stdext::inplace_function<int()>>` in master, by whatever means necessary (I strongly suggest "by means of fixing #125"), and delay merging this PR until it can be merged onto a solid foundation.

Voultapher: Sounds reasonable, I'll get on it.

Voultapher:
rebase @Quuxplusone ping

SG14/inplace_function.h
```diff
@@ -190,15 +223,13 @@ class inplace_function<R(Args...), Capacity, Alignment>
     template<
         typename T,
         typename C = std::decay_t<T>,
-        typename = std::enable_if_t<!inplace_function_detail::is_inplace_function<C>::value>
+        typename = std::enable_if_t<
+            !inplace_function_detail::is_inplace_function<C>::value
+            and inplace_function_detail::is_invocable_r<R, C, Args...>::value
```
Quuxplusone: For MSVC compatibility, this `and` needs to be `&&`. (MSVC doesn't support the English spellings of operator tokens, for some reason unknown to me.)

SG14_test/inplace_function_test.cpp
```diff
@@ -489,6 +489,90 @@ static void test_return_by_move()
...
+// > Determines whether Fn can be invoked with the arguments ArgTypes...
+// > to yield a result that is convertible to R.
+//
+// void is treated specially because a functions return value can be ignored.
+static_assert(is_invocable_r<void, C_Int2, int&>::value, "");
+static_assert(is_invocable_r<const void, C_Int2, int&>::value, "");
+
+// Regression tests for both is_invocable and is_convertible.
+static_assert(is_invocable_r<const int&, int()>::value, "");
+static_assert(is_invocable_r<const int&, int(*)()>::value, "");
```
Quuxplusone: I'd like to see lines 508–543 moved into a function scope, say, `test_is_invocable` (and then called from line 663-ish). That way the typedefs and the `using`-declaration won't leak out into the rest of the file.

SG14_test/inplace_function_test.cpp
```diff
@@ -489,6 +489,90 @@ static void test_return_by_move()
     assert(InstrumentedCopyConstructor::moves == 1);
 }
 
+struct NoDefaultCtor
+{
+    int val;
+    explicit NoDefaultCtor(int v) : val{v} {}
+};
+
+int overloaded_function(stdext::inplace_function<int()>&& exec)
```
Quuxplusone: Every function added in this PR should be `static`. Adding `namespace {` ... `}` around the definition of `struct NoDefaultCtor` is optional IMHO but still probably a good idea.

SG14_test/inplace_function_test.cpp
```diff
@@ -489,6 +489,90 @@ static void test_return_by_move()
...
+
+void test_overloading()
+{
+    EXPECT_EQ(overloaded_function([]() -> int { return 3; }), 3);
+    EXPECT_EQ(overloaded_function([](int arg) -> int { return arg; }), 42);
+
+    using std::is_convertible;
+    using stdext::inplace_function;
+
+    const auto a = []() -> int { return 3; };
```
Quuxplusone: Personally I'd make at least one of a,b,c,d non-`const`, just to test convertibility from non-const prvalues.

SG14/inplace_function.h
Voultapher: Ahh, right forgot that, they do support it but only if you include some extra header, iirc they claim it's for backward compatibility reasons.

SG14_test/inplace_function_test.cpp
Voultapher: Resolved by #157

Voultapher:
friendly ping @Quuxplusone

Quuxplusone:
Merged as a53e9e6122d3b892cd742dc03ab313353c8a3fc4 — thanks for your hard work!

I believe I found one more minor issue which I fixed in 98591c4b9990f95c4978d5ba079f58ae16352feb.

Issue #150, opened 2019-02-04
[inplace_function] Should [](){return 42;} be convertible to inplace_function<void()>?


Quuxplusone:
Related to #149.

```
static void test_void_returning_function()
{
    stdext::inplace_function<void()> f = []() { return 42; };
    f = []() { return 42; };
    f();
}
```

The above test case currently does not compile. Clang's error message is:

```
In file included from inplace_function_test.cpp:2:
../SG14/inplace_function.h:105:15: error: void block should not return a value
            { return (*static_cast<C*>(storage_ptr))(
              ^
../SG14/inplace_function.h:211:31: note: in instantiation of function template specialization 'stdext::inplace_function_detail::vtable<void>::vtable<(lambda
      at inplace_function_test.cpp:412:42)>' requested here
        static const vtable_t vt{inplace_function_detail::wrapper<C>{}};
                              ^
inplace_function_test.cpp:412:42: note: in instantiation of function template specialization 'stdext::inplace_function<void (), 32,
      16>::inplace_function<(lambda at inplace_function_test.cpp:412:42), (lambda at inplace_function_test.cpp:412:42), void>' requested here
    stdext::inplace_function<void()> f = []() { return 42; };
                                         ^
```

If we fix #149, then we must make a design decision here: should this code compile (as it does with `std::function`), or should it fail to compile (because returning `42` from a function with signature `void(int)` usually indicates a programming error)?  https://quuxplusone.github.io/blog/2019/01/06/hyper-function/ is related.

Quuxplusone:
Survey of existing practice (that is, of people who provide a single-header implementation, because I'm lazy): https://godbolt.org/z/PGx6JA

`std::function` happily permits converting `int(*)()` to `function<void()>`.

`boost::function` happily permits converting `int(*)()` to `function<void()>`.

https://github.com/TartanLlama/function_ref `tl::function_ref` consistently forbids converting `int(*)()` to `function<void()>`.

https://github.com/Naios/function2 `fu2::function` claims to permit converting `int(*)()` to `function<void()>`, but if you try it, you get a hard error from the guts of the vtable.

After #155, `inplace_function` claims to permit converting `int(*)()` to `function<void()>`, but if you try it, you get a hard error from the guts of the vtable.

I'd be interested to learn what happens with `folly::function`; with HPX `unique_function`; and/or with MongoDB `unique_function`. But they're not single-header Godbolt-friendly, so I didn't really try hard.

Issue #158, opened 2019-04-22
[inplace_function] Const-callability means thread-safety


Quuxplusone:
Right now, `inplace_function` permits

    auto lam = [i=0]() mutable { return ++i; };
    const stdext::inplace_function<int()> f(lam);
    f();

That is, if I have a reference to a `const inplace_function`, I can call its `operator()`, which can mutate its internal state, even though I've promised not to modify it!

I believe we ought to follow the lead of Folly::Function, fu2::function, etc., and split `inplace_function` into two kinds of specialization:

- `inplace_function<R(As...)>` is constructible from either const-callable or non-const-callable lambdas, and provides only `operator()` (non-const)

- `inplace_function<R(As...) const>` is constructible only from const-callable lambdas, and provides only `operator() const`

In this paradigm, you would not be permitted to write

```
int one(const inplace_function<int(int)>& f) {
    return f(1) + f(2);
}
int two() {
    auto mutlambda = [i = 0](int j) mutable { return i += j; };
    return one(mutlambda);
}
```

The call to `f(1)` would fail to compile because `inplace_function<int(int)>::operator()(int)` is not const-qualified. Instead, you would have to write either

```
int one(const inplace_function<int(int) const>& f) {
    return f(1) + f(2);
}
int two() {
    int i = 0;
    auto nonmutlambda = [&i](int j) { return i += j; };
    return one(nonmutlambda);
}
```

or

```
int one(const inplace_function<int(int) const>& f) {
    return f(1) + f(2);
}
int two() {
    auto mutlambda = [i = 0](int j) mutable { return i += j; };
    return one(std::ref(mutlambda));
}
```
or
```
int one(inplace_function<int(int)> f) {
    return f(1) + f(2);
}
int two() {
    auto mutlambda = [i = 0](int j) mutable { return i += j; };
    return one(mutlambda);
}
```
depending on the semantics you want.

Voultapher:
> Const-callability means thread-safety
This title seems misleading and dangerous, not following const correctness quickly means thread hostile behavior, the inverse is not automatically true.

For once we have the chance to choose the right default in `const` and make mutability opt in.
[...]

PR #159, opened 2019-05-01
Should `int(*)()` be convertible to `inplace_function<void()>`?

Quuxplusone:
Before this patch, we reported that `is_convertible<int(*)(), inplace_function<void()>>`,
but if you actually tried to do it, you got a compiler error down in the guts
of `inplace_function`.

After this patch, we let you do it.

However, see issue #150: I think anyone who relies on this behavior is probably
doing so unintentionally (and their code is probably wrong). Therefore, maybe we
should change our SFINAE so that `is_convertible<int(*)(), inplace_function<void()>>`
would be `false`.


p-groarke:
@Quuxplusone I would personally prefer a fail. I am not a fan of silent issues, and this is clearly a programmer error. However, it may be necessary to conform to `std::function`. Would it be possible to add a custom warning for this?

One could use `[[deprecated("reason")]](C++14)` for it, though not very elegant and certainly not the correct warning type. Might be a stop gap while this is a library. I'm guessing here you can tag dispatch to a deprecated function.

Quuxplusone:
> Would it be possible to add a custom warning for this?

You _can_ use `[[deprecated("reason")]]` as you propose — https://godbolt.org/z/cXpP8M — but as a usability matter, I wouldn't. Most codebases rightly have no notion of "warning"; either an issue is important enough that you make it an error (`-Werror`) and fix it, or it's a non-issue and so you suppress it. Nobody would run long-term with unfixed warnings in their build. So adding such a warning deliberately to `inplace_function` would just force users to choose between avoiding the conversion (i.e. the same outcome as if we'd made it a hard error in the first place), or adding `-Wno-deprecated-declarations` to their build flags (i.e. a worse outcome than the status quo).

Ah! But we _could_ just put the conversion under a compile-time flag like `#ifdef INPLACE_FUNCTION_INVOCABLE_AS_VOID`! That would be pretty clean, I think, unless I'm missing something.

Voultapher:
Hmm, definitely prefer compile-time flag over warning. Afaik that wouldn't be standardizable in this fashion though.

In light of other design changes wanted for inplace_function, I'd recommend to close this hole early and avoid users becoming reliant on it. Imo the advantages of sticking with the existing std::function design and with that its known problems, like this and #158, do not outweigh the disadvantages.

My vote is for never allowing this conversion.

PR #161, opened 2019-05-19
slot_map : Fix generation skipping.

p-groarke:
Slot generation was being incremented both on slot free and on element
insertion. Effectively halving key generation space.

### DIFF
--- a/SG14_test/slot_map_test.cpp
+++ b/SG14_test/slot_map_test.cpp
@@ -1,12 +1,12 @@
 #include "SG14_test.h"
 #include "slot_map.h"
-#include <assert.h>
-#include <inttypes.h>
 #include <algorithm>
+#include <assert.h>
 #include <deque>
 #include <forward_list>
-#include <list>
+#include <inttypes.h>
 #include <iterator>
+#include <list>
 #include <memory>
 #include <random>
 #include <type_traits>
@@ -446,6 +446,32 @@ void BoundsCheckingTest()
     assert(cit == sm.end());
 }
 
+template<class SM>
+void GenerationTest()
+{
+    using T = typename SM::mapped_type;
+    SM sm;
+    auto k = sm.emplace(Monad<T>::from_value(0));
+    int original_cap = int(sm.slot_count());
+    for (int i=1; i < original_cap; ++i) {
+        sm.emplace(Monad<T>::from_value(i));
+    }
+    assert(sm.size() == sm.slot_count());
+
+    sm.erase(k);
+    auto k2 = sm.emplace(Monad<T>::from_value(1));
+#if __cplusplus >= 201701L
+    auto [idx, gen] = k;
+    auto [idx2, gen2] = k2;
+    assert(idx2 == idx);
+    assert(gen2 == gen + 1);
+#else
+    using std::get;
+    assert(get<0>(k2) == get<0>(k));
+    assert(get<1>(k2) == get<1>(k) + 1);
+#endif
+}
+
 void sg14_test::slot_map_test()
 {
     TypedefTests();
@@ -460,6 +486,7 @@ void sg14_test::slot_map_test()
     EraseRangeTest<slot_map_1>();
     ReserveTest<slot_map_1>();
     VerifyCapacityExists<slot_map_1>(true);
+    GenerationTest<slot_map_1>();
 
     // Test slot_map with a custom key type (C++14 destructuring).
     using slot_map_2 = stdext::slot_map<unsigned long, TestKey::key_16_8_t>;
@@ -471,6 +498,7 @@ void sg14_test::slot_map_test()
     EraseRangeTest<slot_map_2>();
     ReserveTest<slot_map_2>();
     VerifyCapacityExists<slot_map_2>(true);
+    GenerationTest<slot_map_2>();
 
 #if __cplusplus >= 201703L
     // Test slot_map with a custom key type (C++17 destructuring).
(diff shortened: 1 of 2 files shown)


SG14_test/slot_map_test.cpp
```diff
@@ -577,3 +609,5 @@ int main()
     sg14_test::slot_map_test();
 }
 #endif
+
+
```
Quuxplusone: (Please remove these blank lines.)

SG14_test/slot_map_test.cpp
```diff
@@ -446,6 +446,31 @@ void BoundsCheckingTest()
...
+    }
+    sm.erase(sm.begin(), sm.end());
+
+    for (size_t i = 0; i < num_iter; ++i) {
+#if __cplusplus >= 201703L
+        auto[idx, gen] = sm.insert({});
+        assert(gen == 1);
+#else
+        using std::get;
+        assert(get<1>(sm.insert({})) == 1);
```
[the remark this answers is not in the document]
p-groarke: It is required. See (your own code?) in `slot_map.h` at line 75,76,77,78 for more examples.

SG14_test/slot_map_test.cpp
```diff
@@ -446,6 +446,31 @@ void BoundsCheckingTest()
     assert(cit == sm.end());
 }
 
+template<class SM>
+void GenerationTest()
+{
+    using T = typename SM::mapped_type;
+    constexpr size_t num_iter = 10;
```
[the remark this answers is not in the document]
p-groarke: What you propose doesn't test the bug fix, though it might be a good test on its own. I've updated the test to do a little more, but there isn't much to test. Just make sure the generations are sequential and aren't skipping values.

SG14_test/slot_map_test.cpp
```diff
@@ -446,6 +446,25 @@ void BoundsCheckingTest()
...
+void GenerationTest()
+{
+    using T = typename SM::mapped_type;
+    SM sm;
+
+    for (int i = 0; i < 10; ++i) {
+        auto k = sm.insert(Monad<T>::from_value(i));
+#if __cplusplus >= 201703L
+        auto[idx, gen] = k;
+        assert(int(gen) == i);
```
Quuxplusone: I feel like this is testing specifically behavior that we _don't_ want (even though my code currently has this behavior). If we change the free-list from FIFO to LIFO, then this test rightly fails, right?

I wrote a test locally that looks like this:

```
template<class SM>
static void GenerationsDontSkipTest()
{
    using T = typename SM::mapped_type;
    SM sm;
    auto k = sm.emplace(Monad<T>::from_value(0));
    auto original_cap = sm.slot_count();
    for (int i=1; i < original_cap; ++i) {
        sm.emplace(Monad<T>::from_value(i));
    }
    assert(sm.size() == sm.slot_count());

    sm.erase(k);
    auto k2 = sm.emplace(Monad<T>::from_value(1));
#if __cplusplus >= 201701L
    auto [idx, gen] = k;
    auto [idx2, gen2] = k2;
    assert(idx2 == idx);
    assert(gen2 == gen + 1);
#else
    using std::get;
    assert(get<0>(k2) == get<0>(k));
    assert(get<1>(k2) == get<1>(k) + 1);
#endif
}
```
That is, we use up all the available indices, and then erase-and-insert one element. We know its index can't change (because it's the only available index), so its generation must increase by 1.

On second thought, I don't think my original idea (writing a test that causes the generation count to wrap around) is terribly good.

SG14_test/slot_map_test.cpp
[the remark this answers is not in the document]
Quuxplusone: Ah, right. `get` here could mean either `TestKey::get` (for the custom key types) or `std::get` (if the key type is `pair`). Okay.

SG14_test/slot_map_test.cpp
p-groarke: In theory, at least how I think one would implement LIFO, my test would still pass. A LIFO would still priorities reusing keys vs. creating new indexes, for memory reasons. So adding and removing 1 element should increment generation by 1, because it is the "last in" key. However, if you choose some more complex memory vs. reuse heuristic, then yes it would fail.

I'll add your test as well, but I do want to keep the original (I like it). More tests the merrier. We can deal with LIFO when that comes around.

p-groarke:
@Quuxplusone Should be gtg. I left the original test in as it is harmless, though I'm ok if you really want to remove it.

SG14_test/slot_map_test.cpp
```diff
@@ -446,6 +446,52 @@ void BoundsCheckingTest()
...
+template<class SM>
+void GenerationTest()
+{
+    using T = typename SM::mapped_type;
+
+    {
+        SM sm;
+
+        for (int i = 0; i < 10; ++i) {
+            auto k = sm.insert(Monad<T>::from_value(i));
```
Quuxplusone: The problem with this test for a LIFO would be that this line (line 458) assumes that "default-construct the container; push_back one element onto it" leaves the container with capacity equal to 1. If the container's capacity started at >=2, or if it preemptively jumped up to >=2 as soon as you put one thing in it, then we'd have more slots at our disposal here than just the one. And so this test is relying on _either_ (what I just described, not happening) _or_ (FIFO behavior for indices).

I believe the test starting on line 472 avoids this issue because it carefully uses up all the already-reserved slots after the first (required) insertion.

p-groarke: I see, that's a good observation. I'll tweak your test a little bit.

p-groarke:
@Quuxplusone ehhhh, what are your commits doing in my PR? lol

Anyhow, I updated the test.

Quuxplusone:
Merged 944379cbb43030190e646a13ab882216e3729834!

p-groarke:
woot

PR #168, opened 2019-09-29
Add installation support

liarokapisv:
These changes provide minimal CMake installation support.
This way consumers can install the headers and reuse accross projects without having to
copy them in each project independently. 

With these changes it will also be possible to add vcpkg and conan support with minimal efford allowing teams that depend on these package managers to more easily use the library.

A note on the following removed lines:

```
-file(GLOB_RECURSE INCLUDES "${SG14_INCLUDE_DIRECTORY}/*.hpp" "${SG14_INCLUDE_DIRECTORY}/*.h")
-target_include_directories(${PROJECT_NAME} INTERFACE ${SG14_INCLUDE_DIRECTORY})
-target_sources(${PROJECT_NAME} INTERFACE
-       $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-       $<BUILD_INTERFACE:${INCLUDES}>
-)

```
`target_sources` is not needed in interface libraries supplying header files since these will be appended to the target's sources to be built. However these will be ignored since cmake does not recognize files ending with `.h` or `.hpp` as source files when building. I also assume that the `target_include_directories` line was simply forgotten since there is a proper invocation a bit later in the script. This one is wrong since it references the source tree in the INTERFACE properties.

Another thing is that the current support requires consumer packages to prefix the headers with SG14 when including. This is good practice but I also allowed the tests be able to include the headers directly.

Any feedback is welcome!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ link_directories(${CMAKE_CURRENT_BINARY_DIR}/lib) # For future use with testing 
 ##
 add_library(${PROJECT_NAME} INTERFACE)
 
+include(GNUInstallDirs)
 
 target_include_directories(${PROJECT_NAME} INTERFACE
 	$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
@@ -79,13 +80,11 @@ endif()
 
 install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}_targets)
 
-include(GNUInstallDirs)
-
 install(EXPORT ${PROJECT_NAME}_targets
     NAMESPACE ${PROJECT_NAME}::
     FILE ${PROJECT_NAME}-config.cmake
     DESTINATION "${CMAKE_INSTALL_DATADIR}/cmake/${PROJECT_NAME}"
 )
 
-install(DIRECTORY "${PROJECT_SOURCE_DIR}/SG14" DESTINATION include)
+install(DIRECTORY "${SG14_INCLUDE_DIRECTORY}" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,13 +28,9 @@ link_directories(${CMAKE_CURRENT_BINARY_DIR}/lib) # For future use with testing 
 ##
 # Project
 ##
-file(GLOB_RECURSE INCLUDES "${SG14_INCLUDE_DIRECTORY}/*.hpp" "${SG14_INCLUDE_DIRECTORY}/*.h")
 add_library(${PROJECT_NAME} INTERFACE)
-target_include_directories(${PROJECT_NAME} INTERFACE ${SG14_INCLUDE_DIRECTORY})
-target_sources(${PROJECT_NAME} INTERFACE
-	$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
-	$<BUILD_INTERFACE:${INCLUDES}>
-)
+
+
 target_include_directories(${PROJECT_NAME} INTERFACE
 	$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 	$<BUILD_INTERFACE:${SG14_INCLUDE_DIRECTORY}>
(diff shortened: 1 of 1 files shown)


p-groarke:
Hi @liarokapisv, thx for the PR. Does this install technique support find_package? If not, I can add full cmake install support.

liarokapisv:
Yes this basically implements `find_package` support. The consumer just uses `find_package(sg14 ...)` and then `target_link_library(lib sg14::sg14)`

p-groarke:
Then 👍 for me.

CMakeLists.txt
```diff
@@ -80,3 +76,16 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
...
+
+include(GNUInstallDirs)
+
+install(EXPORT ${PROJECT_NAME}_targets
+    NAMESPACE ${PROJECT_NAME}::
+    FILE ${PROJECT_NAME}-config.cmake
+    DESTINATION "${CMAKE_INSTALL_DATADIR}/cmake/${PROJECT_NAME}"
+)
+
+install(DIRECTORY "${PROJECT_SOURCE_DIR}/SG14" DESTINATION include)
```
p-groarke: May want to use `${CMAKE_INSTALL_INCLUDEDIR}` install of baking `include`.
Should use `${SG14_INCLUDE_DIRECTORY }` instead of hardcoding `${PROJECT_SOURCE_DIR}/SG14}`.

CMakeLists.txt
```diff
@@ -80,3 +76,16 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
...
 endif()
+
+install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}_targets)
+
+include(GNUInstallDirs)
+
+install(EXPORT ${PROJECT_NAME}_targets
+    NAMESPACE ${PROJECT_NAME}::
+    FILE ${PROJECT_NAME}-config.cmake
+    DESTINATION "${CMAKE_INSTALL_DATADIR}/cmake/${PROJECT_NAME}"
```
p-groarke: Not sure whether this should be `CMAKE_INSTALL_DATADIR` or `CMAKE_INSTALL_LIBDIR`. Thoughts?

liarokapisv: If you search for config files in your system you will see both approaches. `GNUInstallDirs` documents that `LIBDIR` is supposed to be used for "object code libraries" while `DATADIR` should be used for "read-only architecture-independent data". It doesn't really matter since CMake will search in both places, but I think `DATADIR` is the more correct path.

CMakeLists.txt
liarokapisv: Noted.

p-groarke:
Checked out branch, compiles and install without issue. Also tested your branch through `ExternalProject_Add` which clones + calls install in the local prefix. Everything works as expected.

lgtm

Quuxplusone:
These CMake changes are above my pay grade, but I don't mind merging them if there's no objection.

I take it that `${PROJECT_NAME}` is invariably `sg14`? And does installing the project place its header files in `/usr/local/include/sg14`? In the source tree, the header files are located in `SG14/` (all caps). Would it help to move them to `include/sg14/`, and possibly to move the facilities into `namespace sg14` (instead of `namespace stdext`), to reduce the number of names for the same thing we are seemingly accumulating?

liarokapisv:
Yes `${PROJECT_NAME}` is `sg14`. 

For the installation we specify two `target_include_directories` cases. One is when building (eg libraries or in our case the tests) in which case we add the `SG14` folder to the include path allowing tests to use the headers directly (without `SG14` prefix). The other case is after installing in which case it adds the global include directory `${CMAKE_INSTALL_INCLUDEDIR}` (which corresponds to something like `/usr/local/include` in ubuntu for instance) to the consumers' include path. At the end of the script and as part of the installation process we copy the whole `SG14` directory to the global include directory. As consumers only add the global include directory to their include path and not the globally installed `SG14` directory, they need to prefix their headers with `SG14` when including headers. We can always change this of course but it's good practice. 

I think lower-casing the `SG14` directory and moving to something `like include/sg14` is a good idea. In that case, it would also be nice to add the local include folder to `target_include_directories` instead of the final `sg14` folder, unifying the inner-project and consumer's  usage. Both would add the `sg14` prefix when including headers.

I hope this clears up the matter.

PR #171, opened 2020-02-24
CMakeLists.txt - Hide Unit Test related stuff behind option

Jogi7819:
Hide required package THREAD behind option to give the consumer the choice to only add the project interface target without any test dependencies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,14 +1,12 @@
 cmake_minimum_required(VERSION 3.10)
 project(sg14 CXX)
-
-find_package(Threads REQUIRED)
+option(SG14_ENABLE_UNIT_TESTS "Enable unit tests" ON)
 
 # Prefer C++17, downgrade if it isn't available.
 set(CMAKE_CXX_STANDARD_REQUIRED OFF)
 set(CMAKE_CXX_STANDARD 17)
 
 set(SG14_INCLUDE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/SG14")
-set(SG14_TEST_SOURCE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/SG14_test")
 
 # Output binary to predictable location.
 set(BINARY_OUT_DIR ${CMAKE_CURRENT_BINARY_DIR}/bin)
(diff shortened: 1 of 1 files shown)


Quuxplusone:
The offending line was added by @TBBle in b17b9094685e84f9b1d56d18dc6174bd24deff10, and I think I made it obsolete in c5e9751a59e36934acc2b0432cdbb0dc2a931b08 by removing the one (bad) test file that used anything thread-related. Have you tried just reverting b17b9094685e84f9b1d56d18dc6174bd24deff10 and seeing if anything breaks?

Jogi7819:
Will check that today and give a feedback. But still if that commit would be fine I would suggest to hide all test related stuff behind an option that can be set by a top-level project.

Jogi7819:
> 
> 
> The offending line was added by @TBBle in [b17b909](https://github.com/WG21-SG14/SG14/commit/b17b9094685e84f9b1d56d18dc6174bd24deff10), and I think I made it obsolete in [c5e9751](https://github.com/WG21-SG14/SG14/commit/c5e9751a59e36934acc2b0432cdbb0dc2a931b08) by removing the one (bad) test file that used anything thread-related. Have you tried just reverting [b17b909](https://github.com/WG21-SG14/SG14/commit/b17b9094685e84f9b1d56d18dc6174bd24deff10) and seeing if anything breaks?

You are right the package is not required anymore and can be removed. Done in my PR.

CMakeLists.txt
```diff
@@ -40,51 +39,54 @@ target_include_directories(${PROJECT_NAME} INTERFACE
...
+	target_link_libraries(${TEST_NAME} ${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
+	target_include_directories(${TEST_NAME} PRIVATE "${SG14_TEST_SOURCE_DIRECTORY}")
+
+	# Compile options
+	if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+		target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Werror)
+	elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+		target_compile_options(${TEST_NAME} PRIVATE -Wall -Wextra -Werror)
+		set_source_files_properties(${SG14_TEST_SOURCE_DIRECTORY}/plf_colony_test.cpp PROPERTIES
+			COMPILE_FLAGS "-Wno-unused-parameter")
```
Quuxplusone: Could you uncuddle this parenthesis again, so that `git show -b` will have one fewer gratuitous diff?

Quuxplusone (commented):
Seems harmless to me, but also IMHO unnecessary. I don't know much about how we expect this library to be consumed. Does it really make sense for someone to want to consume this library without running its tests? And if they did, would they really do that by wholesale-importing the CMakeLists.txt that we wrote, instead of by just cutting and pasting the single .h file they were interested in?

Jogi7819:
> 
> 
> Seems harmless to me, but also IMHO unnecessary. I don't know much about how we expect this library to be consumed. Does it really make sense for someone to want to consume this library without running its tests? And if they did, would they really do that by wholesale-importing the CMakeLists.txt that we wrote, instead of by just cutting and pasting the single .h file they were interested in?

One aim of modern cmake is to provide everything by target and target specific. Means to me: includes, flags, definitions, etc.
Also I guess to keep external dependencies as small you can, so why adding test related stuff (includes the target itself) if you are not a third-party lib developer. The THREAD dependency is a good example here. It has been required at some point to build the test but it was never required to use the lib itself.

CMakeLists.txt
Jogi7819: Done

PR #172, opened 2020-02-26
Fix warning about no return value in lambda

twam:
The constructor for vtable will save a lambda without return value
to invoke_ptr if SG14_INPLACE_FUNCTION_THROW is used to disable
throwing.
Rename SG14_INPLACE_FUNCTION_THROW to SG14_INPLACE_FUNCTION_NOTHROW and
set invoke_ptr to nullptr_t if throwing is disabled by defining
SG14_INPLACE_FUNCTION_NOTHROW.

### DIFF
--- a/SG14/inplace_function.h
+++ b/SG14/inplace_function.h
@@ -30,10 +30,6 @@
 #include <utility>
 #include <functional>
 
-#ifndef SG14_INPLACE_FUNCTION_THROW
-#define SG14_INPLACE_FUNCTION_THROW(x) throw (x)
-#endif
-
 namespace stdext {
 
 namespace inplace_function_detail {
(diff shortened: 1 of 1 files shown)


Quuxplusone:
FWIW, my original intent with `SG14_INPLACE_FUNCTION_THROW(x)` was that you'd define it to "the thing you want to do instead of throwing": the default is `throw (x)`, but you could define it to `do { log(x); abort(); } while (0)` or to `__builtin_unreachable()` if you wanted those behaviors instead.

You seem to be proposing that there should be only two supported modes — "throw on error" and, um, "don't throw on error"? What is the second mode supposed to do on error? UB?

twam:
UB would be fine for me in that case.
However haven't thought abort `abort();`. That's also fine for me. Thanks!

